PR #8, opened 2016-09-21
feat(node) add node 5.7.0

sleclercq:


rande:
Why do you need this version ? it is not a LTS

sleclercq:
I'll use a previous tag !

PR #13, opened 2016-12-07
Make a lighter Java image by removing caches and locales

fpavageau:


fpavageau:
`docker images` shows that the image went from 989.8 MB to 642.4 MB, by the way.

rande:
Can you split the java space optimisation and put the system wide space optimisation into other dockerfile ?

fpavageau:
Sure. I'll do that tonight, probably.

fpavageau:
Here you go, split in 2 commits:

- one removing extra files and folders from all images
- one removing the downloaded JDK from the cache, and unused parts of the installed JDK: sources, JavaFX, support for GUI tools (VisualVM, Mission Control)

PR #16, opened 2016-12-09
Remove a bit less elements to keep the images functional

fpavageau:
Removing `/var/lib/dpkg/info` prevents `dpkg` and APT from functioning correctly.

Also, `composer` in the PHP images depends on the PostgreSQL lib (`libpq5`) but not as a deb dependency, so in order to keep the `apt-get autoremove` step, the `postgresql-client` is installed.


rande:
@fpavageau libpq5 is still removed ...

```Removing libpq5:amd64 (9.4.9-0+deb8u1) ...```

PR #23, opened 2016-12-15
Pin the version of the installed JDK

fpavageau:


rande:
Travis failed is not related to this commit...

PR #30, opened 2016-12-19
Add an AWS image

fpavageau:
The image provides an environment to manipulate an AWS stack, either with
the shell and awscli, or with a Python script: boto3 and PyYAML are already
installed, more dependencies can be installed via PIP.


rande:
Can you add a quick test in the travis.py file to check if the command are correctly installed ? Thanks

fpavageau:
I've added checks for awscli and the main Python libs.

PR #35, opened 2017-01-03
Update changelog with new image & ci helper version

Bladrak:


Bladrak:
updated @rande

Bladrak:
@rande seems there was a timeout during the travis build, could you retry it? I don't have the permissions.

PR #58, opened 2017-06-29
Add gcc and g++ into Node image

sleclercq:
gcc and g++ are needed by some node dependencies. Having to install them manually on each build creates false negatives because of network issues and is generally inefficient.


fpavageau:
LGTM 👍

PR #71, opened 2017-10-19
fix(dind-aws): add make package and tweak entrypoint to fix daemon co…

mremi:
…nnection

### DIFF
--- a/dind-aws/Dockerfile
+++ b/dind-aws/Dockerfile
@@ -6,7 +6,7 @@ ARG DOCKER_COMPOSE_VERSION
 ARG GLIBC_VERSION=2.23-r3
 
 RUN echo "Install AWS" && \
-    apk add -q --no-cache bash ca-certificates curl git openssh-client python py-pip && \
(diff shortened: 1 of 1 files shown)


dind-aws/Dockerfile
```diff
@@ -25,6 +25,10 @@ RUN echo "Install AWS" && \
     chmod +x /usr/local/bin/docker-compose && \
     docker-compose --version && \
 
+    sed -e "s,\$DOCKER_PORT_2375_TCP,\$EKINO__DOCKER_BUILDBOX_PORT_2375_TCP," \
+        -e "s,export DOCKER_HOST='tcp://docker:2375',export DOCKER_HOST='tcp://ekino__docker-buildbox:2375'," \
+        /usr/local/bin/docker-entrypoint.sh && \
```
fpavageau: To edit the file in-place, you need to call `sed` with the `-i` option as well.

mremi: Oops good catch, forgot it, thx

PR #75, opened 2017-11-16
Add security checker to PHP image

Bladrak:


mremi:
What about the compatibility with PHP 5.3?

Bladrak:
Updated @rande @mremi

PR #82, opened 2017-12-01
Add make to aws image

Bladrak:


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Versions
 ========
 
+2017-12-01
+----------
+
+* Add make to aws image
```
fpavageau: This should actually go in the block below (`2017-11-24`), which should be renamed to `2017-12-DEV` as it corresponds to all the unreleased developments so far. The last release was `2017-11-01`.

Bladrak:
@mremi @fpavageau fixed :)

PR #87, opened 2017-12-18
Add postgres-client to php images

nanofelis:


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Versions
 * Upgrade PHP version: 7.2.0
 * Add rsync to aws image
 * Add zip to aws image
+* Add postgresql-client to php images
```
mremi: s,images,image, ;)

nanofelis: yep I hesitated because they use both syntaxes in the changelog... I chose images cause we create several images from the buimd args. As you wish

nanofelis: ha no, I think they made modification to the 5.3 one as well

nanofelis: I don't know actually x)

mremi: Nop only this Dockerfile is enough, thx

PR #97, opened 2018-03-21
Prevent Yarn from installing Node.js

mr-wildcard:
Yarn automatically installs Node.js by default.

We can skip it by using `--no-install-recommends` as recommended by the [documentation](https://yarnpkg.com/lang/en/docs/install/) : 

> If using nvm you can avoid the node installation by doing:
 `sudo apt-get install --no-install-recommends yarn`

### DIFF
--- a/node/Dockerfile
+++ b/node/Dockerfile
@@ -44,7 +44,7 @@ RUN echo "Starting ..." && \
 
     curl -sSL https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
     echo "deb http://dl.yarnpkg.com/debian/ stable main" >> /etc/apt/sources.list.d/yarn.list && \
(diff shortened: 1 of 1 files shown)


fpavageau:
There's one too many "install" in the commit message:
> Don't let install Yarn install Node.js

should be
> Don't let Yarn install Node.js

or
> Prevent Yarn from installing Node.js

mr-wildcard:
@fpavageau Sorry about that.

fpavageau:
You changed the title of the PR, but not the commit message itself:
```
git commit --amend -m 'Prevent Yarn from installing Node.js'
git push --force-with-lease
```

mr-wildcard:
Damn, I really want Yarn to stop installing Node I guess. I'll rename my commit.

fpavageau:
LGTM 👍

PR #100, opened 2018-04-16
feat : Add Java 9 buildbox

olivierGuillermEkino:
Duplication of java DockerFile to keep java 8 buildbox


.travis.yml
```diff
@@ -12,6 +12,7 @@ matrix:
     - env: LANGUAGE=dind-aws VERSION=1    DOCKER_COMPOSE_VERSION=1.20.1 GLIBC_VERSION=2.23-r3
     - env: LANGUAGE=golang   VERSION=1.9  GLIDE_VERSION=v0.13.1
     - env: LANGUAGE=java     VERSION=8    JAVA_VERSION=8u161-1~webupd8~0
+    - env: LANGUAGE=java9    VERSION=9    JAVA_VERSION=9.0.4-1~webupd8~0
```
mremi: Language should be `java`

olivierGuillermEkino: @mremi @rande Thank you for your feedback. Maybe can you help us to handle this situation.

We need to separate both images because the version number is not the only thing different : java installation folder paths are different too. It would be impossible to know which path to delete since we can't have any logic in the docker file.

For example :
```
/usr/lib/jvm/java-8-oracle/jre/lib/*jfx* \
```
becomes
```
/usr/lib/jvm/java-9-oracle/lib/*jfx* \
```

mremi: You can inject the build arg `VERSION` in [travis.py](https://github.com/ekino/docker-buildbox/blob/master/travis.py#L148) (like ansible), and then use it in the Dockerfile, something like that:
```
rm -rf /var/cache/oracle-jdk${VERSION}-installer && \
if [ “$VERSION” = “8"]; then \
   rm -rf ...; \
else \
   rm -rf ...;\
fi
```

rande:
```
rm -rf /usr/lib/jvm/java-*-oracle/lib/*jfx* \
rm -rf /usr/lib/jvm/java-*-oracle/jre/lib/*jfx* \
```

will work too.

rande:
@olivierGuillermEkino can you solve conflicts ?

PR #102, opened 2018-04-19
Add tests on PHP images

mremi:


rande:
@mremi maybe you can use the tests from symfony

mremi:
@rande if you're talking about https://github.com/symfony/requirements-checker, we can't because of multiple reasons:
* the required PHP version is 5.5.9 at least, but we still support 5.3
* the intl extension is optional, but required in our projects
* the lib expects `composer install` was run before

We could extend it when we won't support 5.3 anymore.

PR #104, opened 2018-04-27
feat(anisble-container): migrate to ansible container

rande:


rande:
@mremi @axelpavageau A small preview of ansible container.

rande:
Ansible container is deprecated from the vendor

PR #111, opened 2018-06-29
feat(php): add Blackfire probe & client

mremi:


Bladrak:
Maybe update the readme as well, explaining quickly how to use it?

mremi:
done

PR #114, opened 2018-07-09
feat(java): migrate java buildbox to openjdk and add jdk 10

Crow-EH:
* build java buildbox from openjdk slim (debian)
* migrate java buildbox to Docker template
* add jdk 10 version
* upgrade react native's buildbox java version to `8u171-1~webupd8~1`

For Information, here are the images informations (with openjdk bases too for comparison) :

```
docker images             
REPOSITORY              TAG                 IMAGE ID            CREATED              SIZE
java10                  latest              3ecd4bff30c4        About a minute ago   595MB
java8                   latest              622bc4077f7d        4 minutes ago        467MB
openjdk                 10.0.1-jdk-slim     2283fee3d3bd        6 days ago           373MB
openjdk                 8u171-jdk-slim      5743387a814e        6 days ago           244MB
ekino/docker-buildbox   java8-2018.06.14    9ceeb231b25a        3 weeks ago          535MB
```
Java 8 is a bit lighter since we now build from openjdk slim (slim debian) VS non-slim debian.
Java 10 is a bit larger, but it's because jdk 10 is larger in the first place.


README.md
```diff
@@ -83,4 +83,14 @@ The ``travis.py`` script try to be clever:
 
 It is possible to build local image for testing with the following command:
 
-    SONARSCANNER_VERSION=3.0.3.778 VERSION=3.0 python travis.py --language sonar --pull-request=true
\ No newline at end of file
+mac :
+
+    SONARSCANNER_VERSION=3.0.3.778 VERSION=3.0 python travis.py --language sonar --pull-request=true
```
mremi: You mean this does not work on Linux??

java/cleanJdk8.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+if [ "$1" == "8" ]; then
```
mremi: Why don't make that in the Dockerfile?

Crow-EH: Because the multiline if else didn't seem to work in the docker file's `RUN`.
I find it cleaner than running a single liner

README.md
Crow-EH: my bad, it does work. I'll correct it

Crow-EH: Done. I replaced the sonar example with the java example since it shows more env variables

java/cleanJdk8.sh
mremi: This should work:
```
    if [ "$JAVA_MAJOR_VERSION" == "8" ]; then \
        echo "Removing unnecessary JDK 8 binaries and libraries ..."; \
        rm -rf /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/javaws \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/*javafx* \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/*jfx* \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libdecora_sse.so \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libfxplugins.so \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libglass.so \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libgstreamer-lite.so \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libjavafx*.so \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libjfx*.so \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libprism_*.so \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/deploy* \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/desktop \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/ext/jfxrt.jar \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/javaws.jar \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/plugin.jar \
               /usr/lib/jvm/java-8-openjdk-amd64/jre/plugin \
               /usr/lib/jvm/java-8-openjdk-amd64/lib/*javafx*; \
    fi && \

    export PATH=$PATH:/docker-java-home && \

    echo "Done!"
```

Crow-EH: Done.
Turns out the main issue with the `if .. else` in Dockerfile was that I had to use `=` instead of `==` since it's runing with `sh`.

Crow-EH:
I upgraded the java version in the React Native build

PR #117, opened 2018-07-26
Add an up to date mime types definition file for the images that use AWS CLI

GuillaumeAmat:
The mime type of the files that are pushed in an AWS storage (and the `Content-Type` HTTP header sent for them) is defined by the [AWS CLI](https://github.com/aws/aws-cli) tool.

You can see [there](https://github.com/aws/aws-cli/blob/5d349108c974d80b5b1a634e8a7433556a31e605/awscli/customizations/s3/utils.py#L294) that the AWS CLI uses the [mimetypes](https://docs.python.org/2/library/mimetypes.html) core Python module to guess the mime types, based on the sender `/etc/mime.types` file.

As the buildboxes use Debian as their base OS, the provider of that file is the [mime-support](https://salsa.debian.org/debian/mime-support) Debian package. But in the current stable Debian version, the more up to date version of the package is 3.60, and the definition file is incomplete (eg: the woff2 mime type is missing).

So, in order to update the definition file and to prevent further missing mime types, I propose to always fetch the last `master` version of the mime-support Debian package.

### DIFF
--- a/aws/Dockerfile
+++ b/aws/Dockerfile
@@ -12,4 +12,7 @@ RUN echo "Install AWS" && \
     echo "Install CI Helper" && \
     curl -sSL https://github.com/rande/gitlab-ci-helper/releases/download/${CI_HELPER_VERSION}/alpine-amd64-gitlab-ci-helper -o /usr/bin/ci-helper && \
     chmod 755 /usr/bin/ci-helper && \
-    echo "Done install CI Helper"
+    echo "Done install CI Helper" && \
+
+    echo "Adding an up to date mime-types definition file" && \
+    curl -sSL https://salsa.debian.org/debian/mime-support/raw/master/mime.types -o /etc/mime.types
--- a/dind-aws/Dockerfile
+++ b/dind-aws/Dockerfile
@@ -25,6 +25,9 @@ RUN echo "Install AWS" && \
     chmod +x /usr/local/bin/docker-compose && \
     docker-compose --version && \
 
+    echo "Adding an up to date mime-types definition file" && \
+    curl -sSL https://salsa.debian.org/debian/mime-support/raw/master/mime.types -o /etc/mime.types && \
+
     echo "Cleaning files!" && \
     rm -rf /tmp/* /var/cache/apk/* && \
 
--- a/php/5.3/Dockerfile
+++ b/php/5.3/Dockerfile
@@ -28,7 +28,6 @@ RUN echo "Starting ..." && \
     phpenv rehash && \
     curl -sSL https://pear.php.net/go-pear.phar -o /tmp/go-pear.phar && \
     php /tmp/go-pear.phar && \
-    rm /tmp/go-pear.phar && \
     pear config-set php_ini $(php -r 'echo php_ini_loaded_file();') && \
 
     ln -s /home/.phpenv/versions/${PHP_VERSION} /home/.phpenv/versions/current && \
--- a/php/5.3/Dockerfile
+++ b/php/5.3/Dockerfile
@@ -26,6 +26,9 @@ RUN echo "Starting ..." && \
     MAKEFLAGS=' -j8' LDFLAGS=-lstdc++ phpenv install ${PHP_VERSION} && \
     phpenv global ${PHP_VERSION} && \
     phpenv rehash && \
+    curl -sSL http://pear.php.net/go-pear.phar -o /tmp/go-pear.phar && \
+    php /tmp/go-pear.phar && \
+    rm /tmp/go-pear.phar && \
     pear config-set php_ini $(php -r 'echo php_ini_loaded_file();') && \
 
     ln -s /home/.phpenv/versions/${PHP_VERSION} /home/.phpenv/versions/current && \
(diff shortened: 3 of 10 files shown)


aws/Dockerfile
```diff
@@ -12,4 +12,7 @@ RUN echo "Install AWS" && \
     echo "Install CI Helper" && \
     curl -sSL https://github.com/rande/gitlab-ci-helper/releases/download/${CI_HELPER_VERSION}/alpine-amd64-gitlab-ci-helper -o /usr/bin/ci-helper && \
     chmod 755 /usr/bin/ci-helper && \
-    echo "Done install CI Helper"
+    echo "Done install CI Helper" && \
+
+    echo "Adding an up to date mime-types definition file" && \
+    curl -sSL https://salsa.debian.org/debian/mime-support/raw/master/mime.types -o /etc/mime.types && \
```
mremi: Alpine here, not a Debian

dind-aws/Dockerfile
```diff
@@ -25,6 +25,9 @@ RUN echo "Install AWS" && \
     chmod +x /usr/local/bin/docker-compose && \
     docker-compose --version && \
 
+    echo "Adding an up to date mime-types definition file" && \
+    curl -sSL https://salsa.debian.org/debian/mime-support/raw/master/mime.types -o /etc/mime.types && \
```
mremi: Alpine here, not a Debian

php/Dockerfile.tpl
```diff
@@ -75,6 +75,9 @@ zend_extension=opcache.so \n\
     pip install -q -U awscli && \
     echo "Done AWS!" && \
 
+    echo "Adding an up to date mime-types definition file" && \
+    curl -sSL https://salsa.debian.org/debian/mime-support/raw/master/mime.types -o /etc/mime.types && \
```
mremi: Alpine here, not a Debian

aws/Dockerfile
GuillaumeAmat: Sure, but still, there is a problem as all the mime types are not known...

Even in Alpine, the main definition file is `/etc/mime.types` and its format is not related to Debian.

```
/data # python
Python 2.7.14 (default, Feb 22 2018, 22:26:54)
[GCC 5.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import mimetypes
>>> mimetypes.guess_type('aze.woff2')
(None, None)
>>> mimetypes.guess_type('aze.woff')
(None, None)
>>> mimetypes.guess_type('aze.txt')
('text/plain', None)
>>> mimetypes.knownfiles
['/etc/mime.types', '/etc/httpd/mime.types', '/etc/httpd/conf/mime.types', '/etc/apache/mime.types', '/etc/apache2/mime.types', '/usr/local/etc/httpd/conf/mime.types', '/usr/local/lib/netscape/mime.types', '/usr/local/etc/httpd/conf/mime.types', '/usr/local/etc/mime.types']
```

What do you think about it @mremi?

GuillaumeAmat:
@mremi The `php` buildbox CI fails because of `pear` not found. After a quick look I found that the base image ([jolicode/phpenv](https://hub.docker.com/r/jolicode/phpenv)) removed it in the latest version and does not propose old versions on the Docker hub :/

How do you prefer to fix it? I install `pear` in our Dockerfile?

GuillaumeAmat:
I can add the proposed commands from the [official installation documentation](https://pear.php.net/manual/en/installation.getting.php):

```
$ wget http://pear.php.net/go-pear.phar
$ php go-pear.phar
```

GuillaumeAmat:
I added a new commit with the installation of PEAR. Just in case you agreed with that ;)

GuillaumeAmat:
It is not really random to me, it fails every single time I want to build the image. And I did not see that it was removed, it was just an assumption.

php/5.3/Dockerfile
```diff
@@ -26,6 +26,9 @@ RUN echo "Starting ..." && \
     MAKEFLAGS=' -j8' LDFLAGS=-lstdc++ phpenv install ${PHP_VERSION} && \
     phpenv global ${PHP_VERSION} && \
     phpenv rehash && \
+    curl -sSL http://pear.php.net/go-pear.phar -o /tmp/go-pear.phar && \
```
mremi: Can you use https and check pear does not exist before installing it please?

php/5.3/Dockerfile
```diff
@@ -26,6 +26,9 @@ RUN echo "Starting ..." && \
...
+    php /tmp/go-pear.phar && \
+    rm /tmp/go-pear.phar && \
```
mremi: Already done below, can be removed

php/5.3/Dockerfile
GuillaumeAmat: Done

php/5.3/Dockerfile
GuillaumeAmat: Oh right, done!

GuillaumeAmat:
@mremi All the fixes are done and the CI is all green 👍 🚀

php/5.3/Dockerfile
```diff
@@ -26,6 +26,8 @@ RUN echo "Starting ..." && \
     MAKEFLAGS=' -j8' LDFLAGS=-lstdc++ phpenv install ${PHP_VERSION} && \
     phpenv global ${PHP_VERSION} && \
     phpenv rehash && \
+    curl -sSL https://pear.php.net/go-pear.phar -o /tmp/go-pear.phar && \
+    php /tmp/go-pear.phar && \
```
mremi: Can you update the `CHANGELOG.md` please?

GuillaumeAmat: Sure, it is done and pushed

PR #118, opened 2018-07-27
feat(sonarts): add sonarts image containing node, npm, sonar-scanner

saugereau:
Since release of official sonarTs plugin, sonar-scanner need the existence of node during run.
This is not needed for previous versions of sonarTS (https://github.com/pablissimo/SonarTsPlugin).
This new image is based upon the Dockerfile of the node buildbox with the adding of sonar-scanner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
     - env: LANGUAGE=python        VERSION=3.6  PYTHON_VERSION=3.6.5
     - env: LANGUAGE=react-native  VERSION=1
     - env: LANGUAGE=ruby          VERSION=2.5
-    - env: LANGUAGE=sonar         VERSION=3.2  GLIBC_VERSION=2.23-r3 SONARSCANNER_VERSION=3.2.0.1227
+    - env: LANGUAGE=sonar         VERSION=10.7 NODE_VERSION=10.7.0-alpine GLIBC_VERSION=2.23-r3 SONARSCANNER_VERSION=3.2.0.1227
 
 script:
     - travis_wait python travis.py
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Versions
 * Add bash to Python images
 * Add an up to date mime types definition file for the images that use AWS CLI
 * Install PEAR in the PHP 5.3 image
-* Update Alpine Version for sonar image: 3.8
+* Replace source image for sonar
 
 2018-06-14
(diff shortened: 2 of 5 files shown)


sonarts/Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM debian:8.7
+LABEL maintainer="Thomas Rabaix <rabaix@ekino.com>"
```
mremi: Sure? :)

sonarts/Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM debian:8.7
```
mremi: Why don't use alpine?

saugereau: I wanted to be the closest to docker ekino's node image, so I use the same origin and use the already working scripts, but maybe a bad idea ?
Is there a global consensus of usage of alpine on all ekino's docker images ?

saugereau: You're right, I have some doubt at the beginning, but I can do almost the same that it was done for java via a docker template file to allow multiple node's version.

.travis.yml
```diff
@@ -25,6 +25,7 @@ matrix:
     - env: LANGUAGE=react-native  VERSION=1
     - env: LANGUAGE=ruby          VERSION=2.5
     - env: LANGUAGE=sonar         VERSION=3.2  GLIBC_VERSION=2.23-r3 SONARSCANNER_VERSION=3.2.0.1227
+    - env: LANGUAGE=sonarts       VERSION=10.7.0-alpine GLIBC_VERSION=2.23-r3 SONARSCANNER_VERSION=3.2.0.1227
```
mremi: `VERSION` should be `10.7` and you should add `NODE_VERSION=10.7.0-alpine` and use it in `travis.py`

sonarts/Dockerfile.tpl
```diff
@@ -0,0 +1,33 @@
+FROM node:{{VERSION}}
```
mremi: Should be `NODE_VERSION`

mremi:
Can you please rebase and fix the conflicts?

.travis.yml
saugereau: Yes, you're right, I will follow what you did on php dockerfile

rande:
The alternative: https://github.com/ekino/docker-buildbox/pull/120/files ;) So now we can install nodejs 8.11

PR #119, opened 2018-08-10
feat(java) add new buildbox using Sun JDK 6

jbhembise:
Creating a new buildbox to compile old code in Java 6, using a JDK from Sun / Oracle.

A Sun JDK 6 was found on old Debian distro : squeeze (6.0). 

Unfortunatly, this distro use an old version of openssl (0.98) which prevents using "curl" over HTTPS in most cases.


jbhembise:
Creating a buildbox for old code which must be compiled in Java 6 using JDK From Sun or Oracle.

A package for Sun JDK ("sun_java6") was found on a old distro of Linux Debian : Squeeze (6.0).

Unfortunatly, this old OS version could not use property "curl" over SSL as OpenSSL version is 0.98 and doesn't meet new security requirements.

PR #124, opened 2018-09-11
update all version langage and tools

fabienChaillou:


fabienChaillou:
I change all version langage for PHP and update the CHANGELOG.md.

PR #127, opened 2018-09-26
Fixed travis.py error messages

eko:
👋 I was boring in the train, looking at your build box (cause I think I will use something like that) and here is this PR ;)

Hope everything is going well for you guys!


mremi:
Hey @eko 😉thx!
Hope you're well too, see you soon

PR #146, opened 2018-12-17
fix(node): lock rb-inotify version to fix issue with Ruby version

mremi:


mremi:
Even better, just use the official image.
I merge to fix the build.

PR #150, opened 2019-02-14
feat(node): add sonar-scanner

GuillaumeAmat:
Add the sonar-scanner tool to the Node.js based images.


rande:
There is already a base image with sonar-scanner: https://github.com/ekino/docker-buildbox/blob/master/sonar/Dockerfile

It is better to split job in the CI.

GuillaumeAmat:
I should have put more details in the description, sorry. I already use the sonar based image, but the `sonar-scanner` tries to call `node`, which is not included in that image, so the job fails. Hence, I need `sonar-scanner` _and_ `node` in the same image.

@rande told me about installing `sonar-scanner` at every pipeline, using the Node based image. But I tend to dislike installing something every single time rather than installing it for good in... A buildbox 🙂

If we want to cut weight in the Node based image, we should start by using Alpine by the way. If I can have your support for that PR with that, I commit to do it right after the merge.

Deal? 🖐

GuillaumeAmat:
Ping @mremi @rande

mremi:
Seems that I miss some information, why Sonar tries to call `node`?

rande:
I don't want to add a new image with `sonar-scanner` + `node`. In order to add node inside the image you can add in the gitlab job those lines:

This will install done inside the scanner image. This is far from perfect, but it is the current trade off.
 
```
        - sed -i 's/4/8/g' /etc/apk/repositories  && apk --update --upgrade add nodejs
        - apk upgrade  apk-tools
        - apk add npm
        - npm -v
        - npm config set unsafe-perm true
        - npm install -g typescript
```

The entire code sample:

```
test_sonar:
    image: ekino/docker-buildbox:sonar3.0-2017.12.12
    stage: test
    tags: [france, ekino]
    script:
        - sed -i 's/4/8/g' /etc/apk/repositories  && apk --update --upgrade add nodejs
        - apk upgrade  apk-tools
        - apk add npm
        - npm -v
        - npm config set unsafe-perm true
        - npm install -g typescript
        - # INSERT YOUR COMMANDS HERE
        - # ...
```

rande:
@mremi the TypeScript plugin required node to do its works.

GuillaumeAmat:
@mremi Because a part of the analysis is done in the container, hence the need of node when the analysed files are JS/TS based.

@rande Like I previously said, I don't want to install the same thing over and over again. This is non sense to me. Even more when we have a *buildbox* which its purpose is to provide all the required tools to do the-damn-thing. I really don't understand your point.

PR #151, opened 2019-02-26
feat(php): bump security-checker versions

fancyweb:
cf https://github.com/sensiolabs/security-checker/issues/149


fancyweb:
@mremi Tests are failing for PHP 5.6 because the security-checker PHAR dependencies (symfony/console for example) are PHP 7.1 +. Where did you find the link to download the PHAR directly? Is there one made with lowest dependencies or should we bundle it ourselves?

mremi:
I think it's time to remove the support for PHP 5.6, legacy projects could still use existing tags.

BTW, could you please switch from http://get.sensiolabs.org to https://get.sensiolabs.org?

Thx

(link to phar is poor-documented [here](https://github.com/sensiolabs/security-checker#usage))

mremi:
@fancyweb can you have a look to the CI please?

PR #153, opened 2019-03-14
Added graphviz to java buildbox

Tetradeus:
Will be used for schema spy for instance

### DIFF
--- a/java/Dockerfile.tpl
+++ b/java/Dockerfile.tpl
@@ -44,6 +44,10 @@ RUN echo "Starting ..." && \
     apt-get -qq -y install maven && \
     echo "Done Install Maven!" && \
 
+    echo "Install graphviz" && \
+    apt-get -qq -y install graphviz && \
+    echo "Done Install graphviz!" && \
+
     echo "Adding an up to date mime-types definition file" && \
     curl -sSL https://salsa.debian.org/debian/mime-support/raw/master/mime.types -o /etc/mime.types && \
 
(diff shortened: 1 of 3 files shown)


java/Dockerfile.tpl
```diff
@@ -44,6 +44,22 @@ RUN echo "Starting ..." && \
...
+        libcdt5 \
+        libcgraph6 \
+        libgd3 \
+        libgvc6 \
+        libgvpr2 \
+        libxaw7 \
+        libxmu6 \
+        libxt6 \
+        libann0 && \
+    curl http://ftp.debian.org/debian/pool/main/g/graphviz/graphviz_2.38.0-17_amd64.deb --output graphviz.deb && \
```
mremi: You should add an argument to manage the version

java/Dockerfile.tpl
```diff
@@ -44,6 +44,22 @@ RUN echo "Starting ..." && \
...
+    dpkg -i graphviz.deb && \
+    rm -f graphviz.deb && \
+    echo "Done Install base!" && \
```
mremi: `base` -> `graphviz`

java/Dockerfile.tpl
mremi: And maybe use the Debian's package?

Tetradeus: Done

java/Dockerfile.tpl
Tetradeus: Done

PR #154, opened 2019-03-27
feat(java): add jq to java buildbox

Crow-EH:


java/Dockerfile.tpl
```diff
@@ -48,6 +48,10 @@ RUN echo "Starting ..." && \
     apt-get -qq -y install graphviz && \
     echo "Done Install graphviz!" && \
 
+    echo "Install jq" && \
+    apt-get -qq -y install jq && \
```
mremi: Should be done in the base install above?

Crow-EH: Yeah, I guess it's not "bug enough" to be considered at the same level thant graphviz or maven.
I'll move it :ok_hand: .
I'll still test it though just to keep things clean.

Crow-EH: Done

mremi:
Failing because of node image which is based on Jessie, but since 20th march Debian has moved the `jessie-updates` repo to archive.debian.org, see https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html.

Do you want to fix it?

Crow-EH:
Yes @mremi, I'll handle it.

Crow-EH:
Turns out the `jessie-updates` (except the security one) has been completly removed and merged in the main repository.

I had to completly override the sources list and of course disable the "valid-until" check, since the packages won't be updated anymore and their validity will expire in the future.

Source : this answer from Stephen Kitt (Debian dev) https://unix.stackexchange.com/a/508728

PR #167, opened 2019-05-28
feat(terraform): add new dockerfile for terraform 0.12.0

julienpierini:


Mallear:
LGTM
@mremi Can you check it an merge please ?

terraform/dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3.6-alpine
+
+LABEL maintainer="Julien Pierini <julien.pierini@ekino.com>"
+
+ARG TERRAFORM_VERSION
+
+RUN apk update
```
mremi: Please merge the `RUN` commands to have only one

mremi: Maybe we should install the [ci-helper](https://github.com/rande/gitlab-ci-helper) as well like other images?

Mallear: Ci-helper isn't necessary in this image. We do not have to manage build dependencies when deploying with Terraform.

mremi:
Why don't use [hashicorp/terraform](https://hub.docker.com/r/hashicorp/terraform)?

JekPi:
because hashicorp/terraform doesn't allow other command than terraform itself (entrypoint on terraform binnary)

mremi:
You can override the default entrypoint through `docker run --entrypoint` or in a docker compose file (see https://docs.docker.com/compose/compose-file/#entrypoint)

JekPi:
Yes but we need python installed in our image and it's easier for us ton start from python-alpine than terraform image

terraform/dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3.6-alpine
+
+LABEL maintainer="Julien Pierini <julien.pierini@ekino.com>"
+
+ARG TERRAFORM_VERSION
+
+RUN apk update \
+    && apk add bash unzip wget ca-certificates \
```
mremi: please sort by name

PR #172, opened 2019-06-19
Refactoring suggestions

Mallear:
Hello there, here is my PR linked to issue #168 .

travis.py
---
I have split travis.py into three files:
- `travis.py` - travis entry point
- `src/config.py` - configuration related functions (environnement variables gathering, image name building etc)
- `src/docker.py` - docker related functions (build, test, login, push)

config.yml
---
Configuration file describing docker images based on the pattern:
```yaml
image_name:
  version:
    cmd_test: [...]
    build_args: [...]
    template_vars: [...]  # if templated Dockerfile
    dockerfile_dir: /path/to/dockerfile  # if Dockerfile's path is not ./<image_name>/Dockerfile (ex: java6 image)
```

.travis.py
---
Update Travis CI configuration to
- fetch dependencies based on Pipenv.lock file before building
- force working with Python3.6 image
- upgrade CI environment to Ubuntu Xenial as Trusty is no more maintained since April 2019.

I have added informations about how to contribute to the project in README.

PR #174, opened 2019-06-24
feat(scoutsuite) : adding a new image using NCC scoutsuite

axelpavageau:
https://github.com/nccgroup/ScoutSuite/


.travis.yml
```diff
@@ -29,6 +29,7 @@ matrix:
     - env: LANGUAGE=react-native VERSION=1
     - env: LANGUAGE=ruby         VERSION=2.6
     - env: LANGUAGE=sonar        VERSION=3.3      GLIBC_VERSION=2.27-r0 SONARSCANNER_VERSION=3.3.0.1492
+    - env: LANGUAGE=scoutsuite   VERSION=5.1.0    PIP_VERSION=19.1.1 PIPENV_VERSION=2018.11.26 SCOUTSUITE_VERSION=5.1.0
```
Crow-EH: I think we should limit the precision of `VERSION` to the minor version (in this case `5.1`), like we do for other images. (well except for ansible it seems 😓)

That way, if someone references the image with `ekino/docker-buildbox:latest-scoutsuite5.1`, the image will have all new patches when pulled again. If not, users will have to upgrade from `ekino/docker-buildbox:latest-scoutsuite5.1.0` to `ekino/docker-buildbox:latest-scoutsuite5.1.1` etc.

It doesn't affect users using more precise / release tags of course, it's juste a choice we give for our images.

**Unless** scoutsuite has the habit of making breaking changes in patch versions of course.

What do you think @axelpavageau ?

axelpavageau: scoutsuite's pretty new (it's a new iteration of a previous project), and so far they've only release major versions.
However, personally, I like setting & using precise versions, but YMMV ...

Crow-EH: Good for me 👍

Crow-EH:
LGTM 👍

.travis.yml
```diff
@@ -29,6 +29,7 @@ matrix:
     - env: LANGUAGE=react-native VERSION=1
     - env: LANGUAGE=ruby         VERSION=2.6
     - env: LANGUAGE=sonar        VERSION=3.3      GLIBC_VERSION=2.27-r0 SONARSCANNER_VERSION=3.3.0.1492
+    - env: LANGUAGE=scoutsuite   VERSION=5.2      PIP_VERSION=19.1.1 PIPENV_VERSION=2018.11.26 SCOUTSUITE_VERSION=5.2.0
```
mremi: please sort by `LANGUAGE`

scoutsuite/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3.7-alpine3.8
+
+ARG PIPENV_VERSION
```
mremi: you have to update `travis.py` to inject those args and test the final image

.travis.yml
Mallear: > I think we should limit the precision of `VERSION` to the minor version (in this case `5.1`), like we do for other images. (well except for ansible it seems 😓)
> 
> That way, if someone references the image with `ekino/docker-buildbox:latest-scoutsuite5.1`, the image will have all new patches when pulled again. If not, users will have to upgrade from `ekino/docker-buildbox:latest-scoutsuite5.1.0` to `ekino/docker-buildbox:latest-scoutsuite5.1.1` etc.
> 
> It doesn't affect users using more precise / release tags of course, it's juste a choice we give for our images.
> 
> **Unless** scoutsuite has the habit of making breaking changes in patch versions of course.
> 
> What do you think @axelpavageau ?

That's why I suggested (in #172) to have one repo per image to follow usual docker image tag practice and allow better comprehension for users and allow us to specify more precise versions

Crow-EH: Yes and we'll be able to have different tagging strategies per repositories if we need to

Crow-EH:
@axelpavageau Can you investigate the CI failure on scoutsuite please?

Crow-EH:
Ok @axelpavageau, I found the issue.

I just noticed that you didn't install `ci-helper` in you Dockerfile - which is fine, not all images need it - but didn't exclude it from the tests.

You need to exclude it in `travis.py` like it's been done for Java 6 and Terraform :
```
if ((language != "java") or (version != "6")) and language != "terraform":
  run_command_exit("docker run %s %s ci-helper version -e" % (run_args, image), "Error with ci-helper installation")
```

This should fix the PR job.

axelpavageau:
@Crow-EH I've fixed the issue with ci-helper, and took the opportunity to add a few tests at the same time.

Crow-EH:
Looks good to me.  With @mremi's review too, I think it can be merged now.

mremi:
For the next time, please squash the commits

Crow-EH:
Sorry about that, I'll pay attention next time.

PR #179, opened 2019-07-24
Create amazonlinux2 with systemd image for Ansible CI jobs

Mallear:
Image needed for ansible roles testing with molecule ci.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -43,6 +43,14 @@ aws:
       PIPENV_VERSION: 2018.10.13
       PY_YAML_VERSION: 3.13
 
+awslnx-systemd:
+  "2":
+    template_vars:
+      VERSION: 2.0.20190508
+    cmd_test:
+      - python --version
+      - pip --version
+
 chrome:
   "1":
     build_args:
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.iml
 
 # Result of template processing
+awslnx-systemd/Dockerfile
 java/Dockerfile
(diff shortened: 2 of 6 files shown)


config.yml
```diff
@@ -43,6 +43,11 @@ aws:
       PIPENV_VERSION: 2018.10.13
       PY_YAML_VERSION: 3.13
 
+awslnx-systemd:
+  "2":
+    template_vars:
+      VERSION: 2.0.20190508
```
Crow-EH: Can't you think about any test that this image does what we want ?

Mallear: This image does not do anything more that having systemd in it. To test if systemd is running, we would need to run it with SYS_ADMIN capability (and some other options). I do not know if Travis allow this and it would require further development of the main script

Crow-EH: Ah, yeah you're right. I just tested and couldn't go anywhere with systemctl.
At least I think you can check that pip and which are installed (maybe by doing a `which pip` 😋?).
We can't really test cronie with a `systemcl is-enabled` though (it was my test).

Mallear:
README updated + added tests for python & pip

awslnx-systemd/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM amazonlinux:2.0.20190508
```
saugereau: I am not sure, but this file isn't it the file generated from your Jinja template ? if yes, maybe isn't it necessary to commit, true ?

Issue #10, opened 2016-12-02
The Java image doesn't actually contain Maven


fpavageau:
It's installed in one step, and removed because of its dependencies right afterwards:
```
The following packages will be REMOVED:
  build-essential* ca-certificates* dpkg-dev* fakeroot* file* fop*
  gsfonts-x11* libaether-java* libcommons-vfs-java* libdoxia-java*
  libfop-java* libice6* libmaven-scm-java* libmaven2-core-java*
  libplexus-containers1.5-java* libsisu-ioc-java* libsm6* libssl-dev*
  libwagon-java* libwagon2-java* libx11-6* libx11-data* libxext6* libxi6*
  libxmuu1* libxt6* libxtst6* manpages* manpages-dev* maven* openssl* patch*
  x11-common* xauth* xfonts-encodings* xfonts-utils* xz-utils* zlib1g-dev*
```
(2nd line from the end)

PR #180, opened 2019-07-24
Create kubectl image

Mallear:
Create kubectl image w/ kubectx, kubens, kube-score for kubernetes deployments

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -126,6 +126,18 @@ java:
       VERSION: 11
       JAVA_VERSION: jdk-11.0.3_7-debian-slim
 
+kubectl:
+  "1.15":
+    cmd_test:
+      - kustomize version
+      - kubectx
+      - kube-score version
+    build_args:
+      KUBECTL_VERSION: 1.15.1
+      KUSTOMIZATION_VERSION: 2.0.3
+      KUBECTX_VERSION: 0.6.3
+      KUBESCORE_VERSION: 0.7.1
+
 node_cmd_test: &node_cmd_test
   - *ci_helper_test
   - node --version
--- a/kubectl/Dockerfile
+++ b/kubectl/Dockerfile
@@ -1,0 +1,29 @@
+FROM alpine:3.10
+LABEL maintainer="Maxime Sibellas <maxime.sibellas@ekino.com>"
+
+ARG KUBECTL_VERSION
+ARG KUBECTX_VERSION
+ARG KUBESCORE_VERSION
+ARG KUSTOMIZATION_VERSION
+
+ENV PATH="${PATH}:/usr/bin"
+
+RUN apk add ca-certificates curl bash wget ncurses && \
+    curl -LO https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl && \
+    chmod +x ./kubectl && \
+    mv ./kubectl /usr/bin/kubectl
+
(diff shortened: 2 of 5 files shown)


config.yml
```diff
@@ -118,6 +118,18 @@ java:
       VERSION: 11
       JAVA_VERSION: jdk-11.0.3_7-debian-slim
 
+kubectl:
+  "1.14":
```
Crow-EH: How about using 1.15, which has been released 6 days ago ?

Mallear: Upgraded to 1.15.1. I don't think it will break CI jobs.

Mallear:
README updated

kubectl/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM alpine:3.10
+LABEL maintainer="Maxime Sibellas <maxime.sibellas@ekino.com>"
+
+ARG KUBECTL_VERSION
```
mremi: Could you please sort by name?

kubectl/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+ARG KUBESCORE_VERSION
+
+ENV PATH="${PATH}:/usr/bin"
+
+RUN apk add ca-certificates curl bash wget ncurses && \
+    curl -LO https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl && \
+    chmod +x ./kubectl && \
+    mv ./kubectl /usr/bin/kubectl
+
+RUN wget https://github.com/kubernetes-sigs/kustomize/releases/download/v${KUSTOMIZATION_VERSION}/kustomize_${KUSTOMIZATION_VERSION}_linux_amd64 && \
```
mremi: Better to have only one `RUN` instruction

Mallear: Why ? This way we can use docker layer sharing properties. As images are downloaded on each gitlab runners, a new version will not require a full download of a new image, only some layers.

kubectl/Dockerfile
Mallear: Done

kubectl/Dockerfile
Crow-EH: [You should minimize your number of layers.
](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers)

It's an old school rule of thumb, and every official image used to be required to follow it too, I don't if it's still the case.

My understanding is that while multiple layers are cool for cacheability (with a caveat,  see below), you produce heavier images and you can't control you image size as efficiently (you can't remove in layer #2 something that you downloaded in layer #1).

And the cache is  not that cool anyway : if you have ten layers and your layer #2 changes, you'll have to rebuild #3 to #10 again.

The new multiple `FROM` is hella cool though :sunglasses:

Mallear: What I understand of that rule is : if you have to compile a tool from sources (download + compile + delete sources), you will have to define all steps inside one `RUN` statement to avoid increasing the weight of the image.

But when you just download binaries, you may want to user docker layers chacheability. And to use it efficiently, you have to order your layer from the least prone to change to the one who will change more often. 

Bonus: multiple layers are way easier to debug when there is a lot of dependencies to install.

fpavageau: Still, it makes no sense to have 2 layers for `kubectx` and `kubens` since they use the same version.

Mallear: Yes, I merged kubectx and kubens RUN statement

PR #181, opened 2019-07-24
Platform.sh CLI image

Mallear:
Docker image with Platform.sh CLI and python3.6.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -209,6 +209,17 @@ php:
     template_vars:
       PHP_VERSION: 7.2.18-cli-alpine
 
+platformsh:
+  "3.44":
+    build_args:
+      PLATFORMSH_CLI_VERSION: "3.44.0"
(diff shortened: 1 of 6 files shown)


config.yml
```diff
@@ -1,5 +1,5 @@
 ---
-docker_hub_namespace: &docker_hub_namespace ekino/docker-buildbox
+docker_hub_namespace: &docker_hub_namespace mallear/docker-buildbox
```
mremi: ?

Mallear: I had to push it to my repo for first use, forgot to revert the change

Mallear:
> Why don't add it to the PHP image?

I need python 3.6 for scripting too

config.yml
Mallear: Change reverted

Mallear:
> What would be the gain of this image when plenty of platformsh-cli exist on dockerhub ?https://hub.docker.com/search?q=platformsh&type=image

Most of them doesn't have python interpreter & when they have, only one "latest" tag is available. This way we control python and tools version & sources.

Issue #182, opened 2019-07-25
Docker image renaming


Mallear:
Hello, there

I'm opening this issue following the merge of my previous PR #172 . 

As discussed in issue #168 and with @Crow-EH in #172, I would like to discuss the possibility to create a Docker Hub repository for each docker image we build.

This way it will be easier to create meaningful tags and find the latest for the image we want, we will be able to give real sementic tags to some image which version hasn't change because it doesn't reflect a tool version (I think of aws or dind-aws image).

I suggest these patterns:

- ekino/<image_name>:< version >-latest/nightly
- ekino/<image_name>:< version >-<tag_date>

for images with multiple versions (java, php, nodes ..)

-  ekino/<image_name>:latest/nigthly
- ekino/<image_name>:<tag_date> 

for images with one version (aws, ansible, chrome ...)

Mallear:
Maybe keeping the buildbox/ci-only notion in the image repository should be good. I'm going with `ekino/<image>-ci` for the repository name.

PR #185, opened 2019-08-02
Change image naming policy to create a repo per image

Mallear:
Here is my code linked to #182 . Format of image renaming is open to discussion.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1,5 +1,5 @@
 ---
-docker_hub_namespace: &docker_hub_namespace ekino/docker-buildbox
+docker_hub_namespace: &docker_hub_namespace ekino
 
 # Common build_vars
 CI_HELPER_VERSION: &CI_HELPER_VERSION 0.0.6
--- a/src/config.py
+++ b/src/config.py
@@ -24,7 +24,7 @@ def load_ci_env(debug):
         info_debug["docker_reg_password"] = "XXX"
         print(">> CI environment configuration: ")
         pp.pprint(info_debug)
-        print('\n')
+        print("\n")
     return build_info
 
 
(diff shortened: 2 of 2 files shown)


Mallear:
That's why I choose to add `-ci` at the end of the name. But still debatable 😄

Crow-EH:
Then why not `gitlab-ci` ? We could make images for other ci systems in the future.

Mallear:
@Crow-EH If we change our CI system, won't we still use docker images ? I don't think images need to change between gitlab, travis or any CI system based on docker images

Crow-EH:
I mean we could offer multiple target systems one day.
Anyway today the readme clearly says `The targeted usage of those images is GitlabCI` so I think our tags should be explicit too.

mremi:
"GitLab" is mentioned in README but the images can be used everywhere, no matter the solution (for local development as well).
I'd use `ekino/dev-{image}:{version}` or still `buildbox` but maybe less explicit?

Crow-EH:
In that case, I'm in favor of `buildbox` too.

Mallear:
Changed to ci-image according to @rande comment in #182

rande:
> I mean we could offer multiple target systems one day.
> Anyway today the readme clearly says `The targeted usage of those images is GitlabCI` so I think our tags should be explicit too.

We can also change the tagline ;) The project has evolved a lot.

Mallear:
I've rebased master on this branch to fix conflicting code. I took the time to make better string building for image name.

Is everyone OK with `ekino/ci-image` pattern or should we discuss more/make a poll to choose image naming convention ?

mremi:
👎 for `ci`, because can be used locally too

flou:
I believe these images are mostly meant for running tests and deployment in a CI environment, but of course they can be used for other uses, any image can be used locally.

In my experience the Python, DinD, Serverless, Terraform images (most of them actually) don't provide much value for local development.
So maybe some images should be prefixed with `dev` but I think `ci` is a better prefix than `dev` or `buildbox`.

GuillaumeAmat:
IMHO, `dev` is too generic, `ci` excludes local usage and `buildbox` is about building things only.

I know that I add some noise to the discussion but, if I need to answer the question, I would pick something like `dev-container-FLAVOR` or `devbox-FLAVOR`.

Mallear:
I have rebased master on this branch.

According to the Slack poll, we will go with `ci-image` pattern.

Mallear:
I'm back. I've rebased new changes from master.

I don't think there is anything more to do for this PR and we are good to go with this change.
I'm available to follow the next release for troubleshooting if needed.

@mremi @rande @Crow-EH @axelpavageau

PR #194, opened 2019-08-29
Create Azure CLI w/ terraform image

Mallear:
Create an image for Azure deployment with Azure CLI and Terraform

### DIFF
--- a/azure/dockerfile
+++ b/azure/dockerfile
@@ -1,0 +1,16 @@
+FROM python:3.7-alpine
+
+LABEL maintainer="Maxime Sibellas <maxime.sibellas@ekino.com>"
+
+ARG TERRAFORM_VERSION
+ARG AZURE_CLI
+
+RUN apk update \
+    && apk add --update -q --no-cache build-base libffi-dev openssl-dev bash ca-certificates unzip wget \
+    && pip install --upgrade --no-binary cryptography pip azure-cli==${AZURE_CLI} \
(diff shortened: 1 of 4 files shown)


config.yml
```diff
@@ -1,5 +1,5 @@
 ---
-docker_hub_namespace: &docker_hub_namespace ekino/docker-buildbox
+docker_hub_namespace: &docker_hub_namespace mallear/docker-buildbox
```
clemstoquart: I think `ekino` was better here :)

Mallear: Oops, fixed :)

clemstoquart: Thanks.

azure/dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN apk update \
+    && apk add --update -q --no-cache bash ca-certificates unzip wget \
+    && pip install --upgrade pip awscli \
+    && wget https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip -O /tmp/terraform.zip \
+    && cd /tmp \
+    && unzip -o terraform.zip \
+    && cd - \
+    && mv /tmp/terraform /usr/bin/ \
+    && chmod +x /usr/bin/terraform \
+    && apk add build-base libffi-dev openssl-dev \
```
mremi: maybe you could group all packages installations

Mallear: done for both apk & pip :)

azure/dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3.6-alpine
```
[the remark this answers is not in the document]
Mallear: Updated

PR #195, opened 2019-09-04
chore(upgrade): upgrade serverless and golang versions

catsoap:


config.yml
```diff
@@ -79,10 +78,10 @@ dind-aws:
       - docker-compose --version
 
 golang:
-  "1.11":
+  "1.13":
```
mremi: You have to update the Dockerfile as well, can you turn it as a template like php for instance?

catsoap: yep sorry, just seen that :p

catsoap: done

mremi: no need to maintain 1.11

config.yml
```diff
@@ -91,6 +90,21 @@ golang:
       - glide --version
       - gin --version
       - modd --version
+    template_vars:
+      GOLANG_VERSION: 1.11
+  "1.13":
```
Crow-EH: You need to reference this version in `.travis.yml`: `matrix.include` (or update the existing one if you end up dropping 1.11) or else Travis won't make a job for it.

config.yml
```diff
@@ -290,11 +304,11 @@ scoutsuite:
       - scout --version
 
 serverless:
-  "1.48":
+  "1.51":
```
Crow-EH: You need to update `env: IMAGE=serverless VERSION=1.48` in `.travis.yml`: `matrix.include`

Crow-EH: It's the matrix Travis use to know which job to run:

image

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ Versions
 * Use only major Node version to name node images
 * Add Node 12
 * Upgrade ScoutSuite to 5.3.3
+- Upgrade Golang version: 1.13
+- Upgrade Serverless version: 1.51
```
Crow-EH: I don't think you need to add this line since the Serverless image has not been released yet.
WDYT @mremi ?

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ Versions
 * Use only major Node version to name node images
 * Add Node 12
 * Upgrade ScoutSuite to 5.3.3
+- Upgrade Golang version: 1.13
```
fpavageau: Can we stick to one style of bullet points, using the same as the lines above?

config.yml
catsoap: thanks, removed

CHANGELOG.md
catsoap: fixed :blush:

PR #197, opened 2019-09-06
feat(serverless): add aws-cli

catsoap:
Hello, here is a small contrib to add aws-cli in serverless image

### DIFF
--- a/serverless/Dockerfile.j2
+++ b/serverless/Dockerfile.j2
@@ -1,4 +1,5 @@
 FROM node:{{NODE_MAJOR}}-alpine
+LABEL maintainer="Loïc Nairi <loic.nairi@ekino.com>"
 
 ARG PIPENV_VERSION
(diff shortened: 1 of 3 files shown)


serverless/Dockerfile.j2
```diff
@@ -13,6 +13,9 @@ RUN echo "Starting ..." && \
     echo "Install basics Python tools" && \
     pip install --upgrade pipenv==${PIPENV_VERSION} && \
     echo "Done basic python tools install" && \
+    echo "Install aws-cli" && \
+    pip install awscli && \
```
mremi: Could you please add a test in config.yml?

mremi: BTW the maintainer is missing, please add him :)

catsoap: done both test and addition of the maintainer

PR #202, opened 2019-09-11
PlatformSH image update

Mallear:
Base platformsh image on php 7.3 alpine & upgrade to 3.47

- Use php 7.3 CLI alpine based image
- Upgrade to platfomsh cli .47

Needed to change image structure as previous one wasn't working without displaying any warning. Some php necessary module weren't installed.


platformsh/Dockerfile
```diff
@@ -8,7 +8,9 @@ ENV PATH="/usr/local/bin/:${PATH}"
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
-RUN apk add --update -q --no-cache bash build-base curl git libffi-dev make openldap-dev openssh-client php7 && \
+RUN apk add --update -q --no-cache bash build-base curl git libffi-dev make openldap-dev openssh-client python3-dev py3-pip && \
```
mremi: Maybe we should just add python 3 (if really required) and platformsh cli to the php image?

Issue #212, opened 2019-09-25
Fix Chrome image build


clemstoquart:
Hi,

For some currently unknown reasons Chrome 77 doesn't work on Alpine Linux.

All commands fail with a cryptic error message :
```
'google-chrome --no-sandbox --version' command failed
Error relocating /usr/lib/chromium/chrome: _ZNSt7__cxx1118basic_stringstreamIcSt11char_traitsIcESaIcEEC1Ev: symbol not found
Error relocating /usr/lib/chromium/chrome: _ZNSt7__cxx1119basic_ostringstreamIcSt11char_traitsIcESaIcEEC1Ev: symbol not found
Error relocating /usr/lib/chromium/chrome: _ZNSt19_Sp_make_shared_tag5_S_eqERKSt9type_info: symbol not found
```

PR #213, opened 2019-09-25
fix: Chrome image

clemstoquart:
Hi,

Fix for https://github.com/ekino/docker-buildbox/issues/212

The Alpine version is now the one provided by the `node:10-alpine` base image. It makes it install Chrome 72 which is outdated but at least it won't broke randomly.


GuillaumeAmat:
Thanks for the update @clemstoquart 👌

PR #218, opened 2019-11-14
chore : Upgrade Scoutsuite version to 5.4.0

axelpavageau:


config.yml
```diff
@@ -298,11 +298,11 @@ ruby:
         - ruby --version
 
 scoutsuite:
-  "5.3":
+  "5.4":
```
mremi: need to update `.travis.yml` too

PR #221, opened 2019-11-14
feat(react-native): upgrade Node/NPM and remove watchman, exposed ports and the custom user

GuillaumeAmat:
Hi there,

I did some small things in the React Native image:

* Upgrade Node.js from v7 to v12 (LTS)
* Remove Watchman
* Remove some exposed ports
* Remove the custom user because of issues with root-owned files (eg: cached dependencies from a previous CI job, based on another Docker image)

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -274,10 +274,11 @@ python:
       PYTHON_VERSION: 3.7.4
 
 react-native:
-  "1":
+  "2":
     build_args:
       CI_HELPER_VERSION: *CI_HELPER_VERSION
       MODD_VERSION: *MODD_VERSION
+      NODE_VERSION: 12.13.0
     test_config:
       cmd:
         - *ci_helper_test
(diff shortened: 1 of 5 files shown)


config.yml
```diff
@@ -272,9 +272,11 @@ python:
 
 react-native:
   "1":
+    build_args:
```
mremi: duplicated

config.yml
```diff
@@ -272,9 +272,11 @@ python:
 
 react-native:
   "1":
```
mremi: should be updated

GuillaumeAmat:
It should be good @mremi

GuillaumeAmat:
@mremi is that ok for you? Thanks

mremi:
Yep, I let you squash the commits and so write a proper commit message

GuillaumeAmat:
👌

GuillaumeAmat:
All good @mremi

PR #231, opened 2019-12-13
chore(dind-aws): bump docker compose

fancyweb:
This new version has a new `--env-file` option that I need to use.


config.yml
```diff
@@ -85,7 +85,7 @@ dind-aws:
   "1":
     build_args:
       CI_HELPER_VERSION: *CI_HELPER_VERSION
-      DOCKER_COMPOSE_VERSION: 1.24.0
+      DOCKER_COMPOSE_VERSION: 1.25.0
```
mremi: please update CHANGELOG.md

Issue #248, opened 2020-02-07
Migrate from alpine to debian + image clean


Mallear:
From discussion started [here](https://github.com/ekino/docker-buildbox/pull/247#issuecomment-583001699), switching for Debian-based images instead of Alpine-based would be a great improvement, at least for all image used for python scripts/app.

Plus it will avoid random failing build caused by Musl lack of compatibility and resiliency.

I will start upgrading these images and clean them (I dont think ci-helper is needed on images only used by ops team).

Mallear:
I think I could need some insight and thoughts on some topics:

- Does anybody use ci-helpers in any of the following images: ansible, arachni, aws, dind-aws, python ?

- I will base images on python:3-buster-slim. I think of managing an image with python 3.6, 3.7 and 3.8. This way every script will have a corresponding image. What are you thoughts ?

clemstoquart:
Hi,

- I don't use ci-helpers on these images
- I don't know what's best here 🤷‍♂️

Crow-EH:
* I don't know
* python buster-slim seems to be a great solution. We you say "every script will have a corresponding image", do you mean you want to build ansible, arachni, aws and dind-aws from ou own python image ?

Mallear:
@Crow-EH I mean I will build a python 3.6, 3.7 and 3.8 image for each. This way we do not force everybody to work on the same python version. Tags will be something like `{tool_version}-p{python_version}-{date}`

Crow-EH:
oh ok, you mean each image will be available in the three python versions. Fine by me then.

Mallear:
Concerning Python 2.7, I'm guessing it is only used to interact with anchore. I'm deleting ci-helper from it. Can you confirm it's ok for you @mremi ?

Concerning other python images, I'm deleting pip and pipenv version fixes. These were needed when there was a non compatible version of pipenv.

Mallear:
Hello,
I've open a PR #250 . I haven't rebased my commit yet as I think this will need to be discuss, at least naming convention. And I may have deleted CI Helper in too many image.

To go further, I would like to propose an alternative to the  dind-aws image, as we cannot control any version of used tool is this image (installing proper python 3.8 version on alpine is a pain).

A workaround to dind image could be the following:
- Sharing the docker socket of the VM running gitlab runners
- Installing docker binary in needed image
- Running docker inside the container linking to the host docker socket.

[...]

Issue #186, opened 2019-08-09
Changelog.md


saugereau:
What about generate changelog on release process (instead of add new line on each development) ?
@mremi , what is the current release process ? Is there any way to use something like https://pypi.org/project/gitchangelog/, to build and commit new Changelog.md during release process (maybe by travis) ?

saugereau:
Current process is : 

- create a branch release
- update changelog.md
- commit 
- merge release branch

The longest part of the current process is rebasing each branch (of feature) to fix the CHANGELOG as you merge each branch one after the other.
As we don’t have a very frequent release process, usually we’ll merge all the feature branches in the days before releasing. Like this week for instance. We have 7 open PRs. So over the next week We need to merge PR1. Rebase PR2 ....
This would be less of an issue if we were merging more frequently

Mallear:
I've found this tool, maybe useful : https://github.com/github-changelog-generator/github-changelog-generator

saugereau:
I try something for this here : https://github.com/saugereau/github-actions-sample, any comment is welcome

Mallear:
The process will be :
- create release/* branch
- merge
- tags master 
?
That sounds very nice :D

clemstoquart:
Thanks @saugereau 

It's ok with me.

saugereau:
> The process will be :
> 
> * create release/* branch
> * merge
> * tags master
>   ?
>   That sounds very nice :D

I confirm @Mallear :)

Crow-EH:
I recommend https://github.com/TriPSs/conventional-changelog-action. It uses https://github.com/conventional-changelog/conventional-changelog under the hood.

They use it themselves, you can see it in their release workflow : https://github.com/TriPSs/conventional-changelog-action/blob/master/.github/workflows/release.yml

axelpavageau:
I'll close this long standing issue, as github now automatically generates changelogs for new releases.

PR #250, opened 2020-02-12
Migrate from Alpine based image to Debian for python related images

Mallear:
This PR aim to migrate python based images from Alpine to Debian for security and easier use. Discussion in #248 .

# Script update

## Templating and python version
 
Each image based on python and may be used to run python script are in fact created 3 times, once for each python version commonly used (3.6 to 3.8). I choosed to name these image with the pattern `{image_version}_p{python_version}_{tag}`.
To define these images in the config file, I've defined a new config part named `multiple_based_version` which should be defined if you want your image being build with different base image versions. It is the parameter checked to determine the image full name. And then define your versions in a `template_args` named `VERSION` which is a liste of version.

## Shorter command line argument

The CLI arguments can now be referenced with `-i` instead of `--image` and `-v` instead of `--version`.

# Updated images by this PR are:

Ansible:
- migration to slim-buster
- delete CI Helper
- Upgrade to 2.8 and 2.9

Arachni:
- deleting CI Helper

AWS:
- migration to slim-buster
- delete CI Helper
- delete NodeJS for CDK, will create a separate image for CDK use

Azure:
- migration to slim-buster

Golang:
- change GO_VERSION variable name for VERSION to be ISO with other templated images

PHP:
- change PHP_VERSION variable name for VERSION to be ISO with other templated images

Python:
- migration to slim-buster
- delete CI Helper

Scoutsuite:
- migration to slim-buster
- delete CI Helper

Serverless:
- migration from node:alpine to python:slim-buster for python runtime control
- delete CI Helper

Terrraform:
- migration to slim-buster
- delete CI Helper


config.yml
```diff
@@ -6,47 +6,53 @@ CI_HELPER_VERSION: &CI_HELPER_VERSION 0.0.6
...
         - pipenv --version
-        - *ci_helper_test
-    build_args:
-      CI_HELPER_VERSION: *CI_HELPER_VERSION
-      NODE_VERSION: 10.16.3-r0
-      PIP_VERSION: 18.1
-      PIPENV_VERSION: 2018.10.13
-      PY_YAML_VERSION: 3.13
+    template_vars:
+      VERSION: *PYTHON_MANAGED_VERSION
```
clemstoquart: Hi, I'm not sure that we need to have the AWS image in 3 different Pyhton version.

Mallear: That's open to discussion, but I don't want to force every one to use the same python version for their scripts. 
Maybe limiting to 3.7 and 3.8 ?

clemstoquart: Or just the last stable release ? People are free to continue using old tags.

saugereau: I am not sure on a use case where a user pull the aws image to, in the same time, build this own app (in python) and use the aws-cli. Maybe the principal requirement must be the own of aws-cli : `require Python 2.7, Python 3.4, or a later version to successfully use the AWS CLI.`
So in my mind only the latest version can be necessary ...

saugereau: And as say @clemstoquart , people are free to continue using old tags

config.yml
```diff
@@ -6,66 +6,76 @@ CI_HELPER_VERSION: &CI_HELPER_VERSION 0.0.6
...
-      VERSION: 2.5.5
-    test_config:
-      cmd:
-        - *ci_helper_test
-        - ansible --version
-        - ansible-playbook --version
-        - python --version
+      ANSIBLE_VERSION: 2.8.8
+    test_config: *ansible_tests
+  "2.9":
```
saugereau: Is there any reason to maintains 2 version of ansible image ? 
The last one's not enough ? (previously we have only one version : 2.5.5)

config.yml
```diff
@@ -126,15 +137,17 @@ java:
       MODD_VERSION: *MODD_VERSION
     test_config: *java_test_config
     template_vars:
-      VERSION: 8
+      VERSION:
+        - 8
       JAVA_VERSION: jdk8u242-b08-debian-slim
   "11":
     build_args:
       CI_HELPER_VERSION: *CI_HELPER_VERSION
```
saugereau: I think that we can delete the ci-helper for the java image, gitlab-ci have improved cache management, so it's not anymore usefull.
I will create a specific issue/improvement for this, so ignore my remark :)

python/Dockerfile.j2
```diff
@@ -1,22 +1,9 @@
...
-    echo "Install CI Helper" && \
-    curl -sSL https://github.com/rande/gitlab-ci-helper/releases/download/${CI_HELPER_VERSION}/alpine-amd64-gitlab-ci-helper -o /usr/bin/ci-helper && \
-    chmod 755 /usr/bin/ci-helper && \
-    echo "Done install CI Helper" && \
     echo "Install basics Python tools" && \
-    pip install --upgrade pipenv==${PIPENV_VERSION} && \
-    if [ "${PYTHON_VERSION:0:1}" = "2" ]; then \
-        pip install anchorecli; \
-    fi && \
+    pip install pipenv awscli boto3 && \
```
saugereau: Not sure, but I think that the content is quiet the same than for aws image, no ?
There is some remove in aws image that are not here ....

config.yml
Mallear: Oh okay, I'll keep only python 3.8 then

config.yml
Mallear: None, that's true. But we need at least one tag of 2.8. I will keep 2.8 for the PR and switch to 2.9 after

config.yml
Mallear: > And as say @clemstoquart , people are free to continue using old tags

I've lost the tag mechanism with the old buildbox ^^'

config.yml
```diff
@@ -6,66 +6,64 @@ CI_HELPER_VERSION: &CI_HELPER_VERSION 0.0.6
...
       ARACHNI_WEB_UI_VERSION: 0.5.12
-      CI_HELPER_VERSION: *CI_HELPER_VERSION
     test_config:
       cmd:
         - arachni --version
-        - *ci_helper_test
 
 aws:
   "1":
+    multiple_based_version: true
```
clemstoquart: Do we really still need that ?

config.yml
```diff
@@ -6,66 +6,64 @@ CI_HELPER_VERSION: &CI_HELPER_VERSION 0.0.6
...
+      VERSION:
+        - 2.0.20190508
     test_config:
       cmd:
         - python --version
         - pip --version
 
 azure:
   "2":
+    multiple_based_version: true
```
clemstoquart: Idem

config.yml
Mallear: Not anymore

config.yml
Mallear: deleted

Mallear:
Commit rebased

config.yml
```diff
@@ -6,47 +6,39 @@ CI_HELPER_VERSION: &CI_HELPER_VERSION 0.0.6
...
         - pipenv --version
-        - *ci_helper_test
-    build_args:
-      CI_HELPER_VERSION: *CI_HELPER_VERSION
-      NODE_VERSION: 10.16.3-r0
-      PIP_VERSION: 18.1
-      PIPENV_VERSION: 2018.10.13
-      PY_YAML_VERSION: 3.13
+    template_vars:
+      VERSION: 3.8
```
Crow-EH: I know this is the base python image's version, but not everybody will. We should name it either `PYTHON_VERSION` to be explicit, or `BASE_IMAGE_VERSION` to be more generic (or something like that).

README.md
```diff
@@ -61,7 +61,8 @@ image_name:
       volume: ... # docker volume if needed, format: localdir:/path/to/mount
       cmd: [...]  # shell commands run to be sure tools are well installed
     build_args: [...]  # If ARG are defined in Dockerfile
-    template_vars: [...]  # If templated Dockerfile
+    template_vars:  # If templated Dockerfile
+      VERSION: # If multiple image with different based image version is created
```
Crow-EH: This is not really mandatory I think : any image is very much free to name its own variables. Unless we decide to find a generic name for the version of the base image of course.

config.yml
Crow-EH: ie: `VERSION` was ok for our java image, since it's java's version. Now it's a completly different thing.

README.md
Crow-EH: IMHO `If the templated Dockerfile describes multiple images with different base images` is a bit better

README.md
```diff
@@ -70,6 +71,8 @@ Do not forget to add an entry in `.travis.yml` too following other image scheme.
 
 **Volume mounting** for test configuration only need the directory name as full local path is build by the script.
 
+**Templating** is used ONLY for based image version, for library version please prefer build args. Base image version(s) should be define in a variable named `VERSION`.
```
Crow-EH: Enforcing best practices is always a good idea :+1:.

WDYT about :
```suggestion
**Templating** must **ONLY** be used for images with multiple base image versions. For any other variables used in a Dockerfile, prefer build args. Base image version(s) should be defined in a variable named `VERSION`.
```
With the last part removed if we don't enforce variables names ?

Crow-EH:
Also, please update the changelog

config.yml
Mallear: Or just read the README about the variable role. Maybe using `BASE_IMAGE_VERSION` should be better tho

README.md
Mallear: The comment about `VERSION` will be deleted, it was used when I wanted to build multiple image with different base version (python 3.6, 3.7 for the same tool version)

Mallear:
> Also, please update the changelog

We will have to discuss about #186 after this PR

README.md
Mallear: Updated 👍

config.yml
Mallear: Switched to BASE_IMAGE_VERSION for every image. More meaningfull, and README updated.

Mallear:
> Also, please update the changelog. Otherwise good job ! 👍

Changelog updated and commit rebased 👍

Crow-EH:
@saugereau @axelpavageau do you have anything to add, or can we merge this ?

PR #254, opened 2020-02-24
CDK/Serverless image

Mallear:
Create an image for CDK and serverless use cases.


Mallear:
I'm currently having some trouble with this new image, still testing.

Mallear:
OK for merge ! 
Need to maintains 3.7 and 3.8 versions as AWS Lambda doesn't support 3.8 yet and serverless need to be run in the same version as the lambda runtime.

May delete the 3.7 after the 1rst release.

clemstoquart:
@Mallear Do we really need another image ?

Do you think it would be possible to use the existing serverless image ?

Mallear:
> @Mallear Do we really need another image ?
> 
> Do you think it would be possible to use the existing serverless image ?

Yep, that's why I renamed serverless image and just deleted serverless as it is managed as a project dependency to have the right version for each project.

clemstoquart:
Ok :)

BTW you should probably use Node 12 instead of 13.

Mallear:
> Ok :)
> 
> BTW you should probably use Node 12 instead of 13.

More reliable ? More secure ? I will update this and rebase 👍

clemstoquart:
> > Ok :)
> > BTW you should probably use Node 12 instead of 13.
> 
> More reliable ? More secure ? I will update this and rebase 👍

Odd Node release are considered as cutting edge version that should only be used if you need the latest features. Staying on the last LTS release is safer.

Mallear:
> > > Ok :)
> > > BTW you should probably use Node 12 instead of 13.
> > 
> > 
> > More reliable ? More secure ? I will update this and rebase 👍
> 
> Odd Node release are considered as cutting edge version that should only be used if you need the latest features. Staying on the last LTS release is safer.

Good to know, thanks :)

clemstoquart:
I don't think the build error is related.

Mallear:
Fixed :D

README.md
```diff
@@ -183,6 +183,12 @@ https://hub.docker.com/r/ekino/ci-python/tags
 
 Contains Python with PIP and PIPENV.
 
+### Pynode
+https://hub.docker.com/r/ekino/ci-pynode/tags
+
+Based on Python image, contains Nodejs 13. Used for serverless and CDK use cases, with tools installed in a
```
clemstoquart:
```suggestion
Based on Python image, contains Nodejs 12. Used for serverless and CDK use cases, with tools installed in a
```

Mallear: Fixed 👍

PR #258, opened 2020-02-27
Add git in AWS Image

Mallear:
We discovered on ekino/ci-aws:latest that we cannot make troposphere diffs as it needs git and git isn't part of debian:buster-slim.


Mallear:
> Can you add a test please.

Done 👍

clemstoquart:
> > Can you add a test please.
> 
> Done 👍

Thanks.

PR #262, opened 2020-03-05
feat(tezos): add tezos image for qa

catsoap:
Hello I propose to add this image for Tezos blockchain

did not add config test because the added programs do not have option to display version, but tested locally, everything ok

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,6 +47,7 @@ jobs:
           - { image: scoutsuite, version: 5.8 }
           - { image: sonar, version: 3.3 }
           - { image: terraform, version: 0.12 }
+          - { image: tezosqa, version: 0.1 }
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


Bladrak:
@catsoap so far so good, what's pending for WIP status?

catsoap:
Hello ! it needs an update because SmartPy url has changed, and maybe we can add truffle@tezos and ligo, I'll do an update later today !

catsoap:
@Bladrak It's ok now

Bladrak:
@catsoap the truffle config thing might be worth documenting within a readme file next to the dockerfile, would you mind adding that?

catsoap:
yep, but I'm sorry, I was wrong, it's not currently possible https://github.com/trufflesuite/truffle/blob/alphaTez/packages/compile-ligo/index.js#L88, will have a look at what we can do

catsoap:
Hi sir, I removed truffle, the image is currently good enough, we can:
- compile smartpy
- compile ligo
- test with pytezos

Bladrak:
Ok, looks good to me, thanks!

PR #265, opened 2020-03-11
IBM: Upgrade provider version to 1.2.4

Mallear:
Upgrade provider to 1.2.4 for fixes and new resources management


Mallear:
Fix CHANGELOG for `Move ibm image to Golang base image and build Null provider` which was listed under wrong release.

PR #271, opened 2020-03-22
feat(ci): build only images with changes

Crow-EH:
WIP ~~and blocked by #239~~

The built-in conditional build on path change only works at workflow level, implying one workflow / image, and there is still no way to share resources between workflows, so I chose to use a custom solution to keep things clean.

The mechanism depends on git diff (against master by default, for push and pull-request events), through a third-party Action, to check if an image has been modified: it works with multiple commits.

Also splits the config by images.

We want some events to trigger a build even without changes: the cron and the release.


rande:
@Crow-EH base images can change too so you still need to ensure build process test those cases.

Mallear:
Seems good for me ! And it's a good idea to split config files, it will be easier to maintains and will result in less merge conflict :) 

Maybe copy/pasting version of CI Helper and modd version in all image using it ? That will create redundancy but may be easier for you than tweaking the diff process one more time.

Concerning release, I think all image should be build and push at each release, with or without changes. This way, the naming pattern is valid for all images.

Crow-EH:
@rande Do you mean the base image (FROM) changing without changing its version?

Won't the fact that the diff will be applied only on PR and push be enough? The cron is there to check everything builds, and the release will build everything too.

Also, don't we have the same problem today? X)

@Malllear for CI helper and modd, checking the base config.yml for changes should work, but will build everything when it changes. I'm cool with that, how about you?

Mallear:
As far as I read the third-party action doc, your PREFIX_FILTER definition will trigger a build if the Dockerfile is changed as well as config.yml. That should be OK for all change possible.

Diff applied only on PR and Push is Ok for me, cron should build and push nightly.

As CI Helper and modd are not upgraded often, we can keep this. But if it makes tricky to configure the CI, I think we should prefer redundancy.

saugereau:
Documentation in https://github.com/technote-space/get-diff-action is not clearly explicit about what happen if you just set PREFIX_FILTER and not SUFFIX_FILTER, but we can assume that it works :)
And I am agree to use this mecanism for push and pull_request events only, as Mallear think,  I think all image should be build and push at each release, with or without changes.

Crow-EH:
I tweaked the watched paths and the condition. Now the build of an image will happen only if :
* Something changed on the path `<image-name>/**` or in the root `config.yml`
* Or if the event is neither a `push` nor a `pull_request`

Do you think I should add those paths to the watch list :
* `.github/workflows/build.yml`: Shouldn't be necessary since an image version change should change its `config.yml` too.
* `image_builder.py` and `src/**`: Do we need to rebuild everything if our build system evolves, or can we wait for an image version upgrade ?

Mallear:
No need for build definiton, it will trigger a new job when you first commit the new image.

For the script, it may be usefull when upgrading the script, to be sure all images build correctly. That's something we can test locally but it's better to run jobs against potential modifications.

Crow-EH:
I've added `image_builder.py` and `src` to the watch list.

**Next question:** Now that we don't have to wait for each image, modified or not, to build (or fail) on pull requests, do you think we should put back the `fail-fast: false` in our build strategy ?

It will allow us to push or release on docker hub even if an unmodified image fails to build.

**Example of behaviour without fail-fast: false**
I update the `java` image:
* My PR only triggers the build for `java`: the build is passing
* The code is merged at a date D1: `node` is failing, let's say because https://dl.yarnpkg.com/debian/pubkey.gpg is temporarily unavailable => `java` will be pushed to docker hub only if it passes before `node` fails
* (Without any error on master) The code is released at a date D2: `ibm` is failing for once again an issue we couldn't foresee => we don't release `java` because of a random issue on `ibm` (or maybe we do because java executed first)

Mallear:
@Crow-EH  Ok, may be useful to setu fail-fast to true :)

Crow-EH:
> @Crow-EH Ok, may be useful to setu fail-fast to true :)

You mean false (deactivating it) right ? :smile:

Mallear:
> > @Crow-EH Ok, may be useful to setu fail-fast to true :)
> 
> You mean false (deactivating it) right ? 😄

Oh yes, totally :D

saugereau:
> put back the `fail-fast: false`

Yes, I think that we can put back the `fail-fast: false`
Another question, what about "Require branches to be up to date before merging an status check" settings ? We can leave it as before, true (because we will not build all images) ?

Crow-EH:
> > put back the `fail-fast: false`
> 
> Yes, I think that we can put back the `fail-fast: false`
> Another question, what about "Require branches to be up to date before merging an status check" settings ? We can leave it as before, true (because we will not build all images) ?

I'm working on a PoC to have a single workflow required : https://github.com/ekino/docker-buildbox/pull/275

Crow-EH:
Just got a good idea from @Mallear.

Currently, I'm using an action for:
* **At the start of each job**: git diff according and exposing an env var if a pattern matches
* **At the start of each following step**: execute only if the env var is truthy (with `if` at step level)

That works great, and it speeds up the builds a lot when we don't have many modified images, but we can still improve it.

We could use an action for:
* **At the start of the workflow (in a dedicated job)**: git diff and expose a kind of searchable structure
* **At the start of each job**: execute only if the pattern was found in the structure (with `if` at job level)

I don't know if I'm making myself clear, but this should improve our overall build speed. :fire: 

I'll find a way to do it in the following days (and test everything again :sweat:),

Crow-EH:
Update: conditional matrix jobs are sadly not possible. The job's `if` is interpreted before the matrix, so you either run everything or nothing (and obviously can't decide to run it bu matching an image name and the diff).

The gate step system I used is working but not that great (all jobs still have to run, which takes time).

I'll work on a better (but not simpler) solution: [We can now build a dynamic matrix with `fromJSON()`](https://github.blog/changelog/2020-04-15-github-actions-new-workflow-features/), and with the way I was planning to reorganize the images configuration (each image has its own configuration file), we do not really need to repeat the available versions in a centralized matrix anyway.

Long story short, I plan to create a first job that'll:
* Git diff
* Lists all image folders: Neither src nor .github
* Evaluates which images need to be in the matrix: Event requires full build || some global file/folder was changed || the image folder was changed
* Generate and output a matrix JSON based on each image's config

Crow-EH:
a bit too old for my tastes, plus I prefer to rethink the way we'll do it before trying again. Closing it.

PR #274, opened 2020-04-02
Adding kubernetes terraform provider and JQ to IBM image and upgrade IBM provider to 1.4.0

Mallear:


Mallear:
Currently testing the image

Mallear:
> Currently testing the image

Tested, ready to merge

Crow-EH:
We also need to remove ibm 1.2 from the mandatory checks, and add ibm 1.3 :disappointed:

Mallear:
> We also need to remove ibm 1.2 from the mandatory checks, and add ibm 1.3 😞

Is that something we need to do manually ?

Crow-EH:
Yup, everytime we change the matrix. Github automatically detects new matrix jobs and run it, but as you can see here, they are not automatically required before merging :
image

And the old jobs are stil required:
image

There's still no built-in way to have one global check on the workflow when you use a matrix.

Crow-EH:
The easiest solution today woul be to stop requiring jobs to pass before merging.

The jobs would still be listed, the merge button would just not be disabled anymore.

Mallear:
> 
> There's still no built-in way to have one global check on the workflow when you use a matrix.

And not using matrix is way harder to setup our builds ?

Crow-EH:
Not using matrix would mean having either :
* 1 workflow / current matrix conf : We would have the same problem anyway
* having 1 big job and delegating the "matrix" usage to python

Another solution would be to create a separate workflow that would wait for the current one to end, check its status with GitHub API, then fail or succeed accordingly.

We could then require this workflow only.

I'll try to PoC it ASAP :)

PR #284, opened 2020-05-05
Add forgotten pynode image in build list

Mallear:


.github/workflows/build.yml
```diff
@@ -34,6 +34,8 @@ jobs:
           - {IMAGE: php, VERSION: 7.3}
           - {IMAGE: php, VERSION: 7.4}
           - {IMAGE: platformsh, VERSION: "3.50"}
+          - {IMAGE: pynode, VERSION: 3.7}
```
Crow-EH, line 37: Can you also fix the sorting while you're at it, please ?
`ibm` should be after `golang`.

Otherwise, LGTM

Mallear: Done

PR #292, opened 2020-05-26
Add kubeval, kustomize and AWS Terraform provider in ibm image

Mallear:


ibm/Dockerfile.j2
```diff
@@ -33,6 +36,11 @@ RUN apt-get update -qq \
     && cd terraform-provider-null-${IBM_NULL_PROVIDER_VERSION} \
     && make build \
     && mv $GOPATH/bin/terraform-provider-null /terraform.d/plugins/terraform-provider-null_v${IBM_NULL_PROVIDER_VERSION} \
+    && wget https://github.com/terraform-providers/terraform-provider-aws/archive/v${AWS_PROVIDER_VERSION}.zip \
```
Crow-EH, line 39: It's becoming a messy wall of text I think. :sweat_smile: 

Can you please split the blocks by intent, and maybe add a comment / title for each block ?

Or even better : use `echo` like we did in `dind` or `java`.

Mallear: Empty lines are deprecated:
```
[WARNING]: Empty continuation lines will become errors in a future release.
```
I will split into multiple RUN statements

Mallear: Ok, split with echo

Mallear:
We have to find something to end with merge conflict on CHANGELOG file ...

ibm/Dockerfile.j2
Crow-EH: Sure, there's not any issue with mutliple RUN statements now that the officials debian (which goland is based on through buildpack) and ubuntu are automatically cleaning after the install : https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105

At least according to the doc : https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

image


Before this feature, multiple RUN would render the cleaning useless, since you can't remove a file created in layer N from layer N+X.

Crow-EH:
Yup, we need to revive #186

ibm/Dockerfile.j2
Mallear: Oh nice, didn't know. I will split into multiple RUN statement then, give me 5min

Mallear: Done

Crow-EH:
LGTM :+1:

PR #295, opened 2020-06-09
chore: upgrade Java 8 to 8u252 and 11 to 11.0.7

clemstoquart:


fpavageau:
I would have merged, but there's a conflict now.

PR #299, opened 2020-06-19
Add IBM Cloud CLI plugin for Functions

Mallear:


ibm/Dockerfile.j2
```diff
@@ -23,6 +23,7 @@ RUN apt-get update -qq \
     && ibmcloud plugin install kubernetes-service \
     && ibmcloud plugin install container-registry \
     && ibmcloud plugin install cloud-databases \
+    && ibmcloud plugin install functions \
```
Crow-EH, line 26: How about testing plugins installation from mow on ?

I know you should be able to test `cloud-functions` with `ibmcloud fn list` [according to the doc](https://cloud.ibm.com/functions/learn/cli), I don't know if it's the same for `functions`.

I think we should guarantee that the plugins are working even when we upgrade ibmcloud's version.

What do you think @Mallear ?

Mallear: As far as the plugin installation succeed, there is no way it won't work.
Furthermore, the only way of testing plugins without the need to login to an existing ibmcloud account (and store API keys and creds as secrets etc) would be to do  `ibmcloud plugin --help"
I can add this as a test command, but I'm not sure this is useful

Crow-EH: Alright, you know it better than me anyway, so I'll follow your advice and approve it as is.
As usual, thank you for your contribution. :+1:

PR #313, opened 2020-07-29
fix(check): no statuses assumed as success

Crow-EH:
WIP

Definitively not due to launch order, still investigating

Issue created on their repo : https://github.com/WyriHaximus/github-action-wait-for-status/issues/21


Crow-EH:
Actually, it turns out out "statuses" are the old, external checks, but "checks" are the github actions checks.

It is working correctly so I'm closing this.

PR #333, opened 2020-09-02
chore(upgrade): upgrade golang version to 1.15.1

catsoap:


Crow-EH (changes requested):
You also need to edit `.github/workflows/build.yml` :
```
- { image: golang, version: 1.14 }
```

PR #336, opened 2020-09-17
chore(golang): add rsync to golang image

catsoap:


config.yml
```diff
@@ -129,6 +129,7 @@ golang:
         - mockgen -version
         - modd --version
         - swagger version
+        - rsync --version
```
mremi, line 132: please sort by command name :)

golang/Dockerfile.j2
```diff
@@ -15,7 +15,7 @@ RUN echo "Starting...\n" && \
 
   echo "Installing AWS CLI..." && \
   apt-get update -q && \
-  apt-get -qq -y install python-pip groff-base && \
+  apt-get -qq -y install python-pip groff-base rsync && \
```
mremi, line 18: same here, please sort

config.yml
catsoap: thanks, fixed

golang/Dockerfile.j2
catsoap: fixed !

PR #338, opened 2020-10-12
feat: upgrade terraform version to 0.13.4 and terragrunt to 0.25.3

julienpierini:


config.yml
```diff
@@ -439,6 +439,21 @@ terraform:
         - git --version
     template_vars:
       BASE_IMAGE_VERSION: 3.8
+  "0.13.4":
```
clemstoquart, line 442: Hello, I think you can override the version `0.13.1` instead of creating a new one.

clemstoquart (changes requested):

config.yml
Mallear: Terraform new versions aren't retro-compatible. We need to maintain multiple minor versions.

Mallear: I think we can stop maitaining 0.12.29. Can you confirm @axelpavageau ?

clemstoquart: Ok 😢

config.yml
```diff
@@ -439,6 +439,21 @@ terraform:
...
+    build_args:
+      TERRAFORM_VERSION: 0.13.4
+      TERRAGRUNT_VERSION: 0.25.3
```
Mallear, line 445: You can upgrade other image's terragrunt version.

clemstoquart (approved):

config.yml
```diff
@@ -439,6 +439,21 @@ terraform:
...
+      TERRAFORM_COMPLIANCE_VERSION: 1.3.2
+    test_config:
+      cmd:
+        - terraform version
+        - terragrunt --version
+        - terraform-compliance --version
+        - python --version
+        - aws --version
+        - git --version
```
Mallear, line 454: Can you use YAML anchors to declare this block of cmd only once ?

PR #361, opened 2021-01-13
feat(php): add php 8.0.0 image, bump composer and php-cs-fixer versions

mazsudo:


mazsudo:
seems almost ready @mremi ;)

PR #383, opened 2021-03-01
chore(php): add pcov in php images

mazsudo:


php/Dockerfile.j2
```diff
@@ -47,6 +47,8 @@ RUN echo "Starting ..." && \
     docker-php-ext-install -j$(getconf _NPROCESSORS_ONLN) bcmath exif gd intl mysqli pcntl pdo_mysql pdo_pgsql pgsql soap sockets zip && \
     pecl install apcu-${APCU_VERSION} && \
     pecl install memcached-${MEMCACHED_VERSION} && \
+    pecl install pcov && \
+    docker-php-ext-enable pcov && \
```
Bladrak, line 51: Maybe also add a test (see https://github.com/ekino/docker-buildbox/blob/master/php/test.php#L144)

Issue #395, opened 2021-04-08
Add percy container


rande:
Objectif: being able to run Percy tests

Feature required: 
- Accept env variables to configure the local agent (TOKEN, etc ...)
- Custom script to parse a sitemap.xml file and iterate over the file to analyse pages; this can help: https://github.com/ekino/rendr/blob/master/examples/gatsby/gatsby-node.js#L33-L65
- Able to start a command to init the test suite

Env Variable:
- SITEMAP_XML="https://domain.com/sitemap.xml" - if set, start the sitemap check, for now just let support basic sitemap (not nested definition)
- PERCY_TOKEN=XXXX - mandatory value
- START_COMMAND=percy.js - if set, start the command to run the Percy tests

Crow-EH:
We still need to document the new image in our readme before closing this issue

PR #397, opened 2021-04-15
WIP: add Percy buildbox.

ekinoben:
See https://github.com/ekino/docker-buildbox/issues/395

Adds a buildbox with [Percy](https://percy.io/) installed. Box includes a custom JS script to process a given sitemap XML file.


percy/sitemap.js
```diff
@@ -0,0 +1,37 @@
...
+  console.log('[INFO] Found %d page references', pageReferences.length);
+  for (const i in pageReferences) {
+
+    const url = parseUrl(pageReferences[i].loc)
+    console.log('[INFO] Processing %s', url.href);
+
+    await PercyScript.run(async (page, percySnapshot) => {
+        await page.goto(url.href);
+        // ensure the page has loaded before capturing a snapshot
+        // await page.waitForSelector("body");
```
rande, line 31: @ekinoben what `waitForSelector` is for?

percy/sitemap.js
```diff
@@ -0,0 +1,37 @@
+const Axios = require("axios")
+const PercyScript = require("@percy/script");
+const parseUrl = require("url-parse")
+const xml = require("xml2json")
+
+async function loadSitemap(url) {
+  const response = await Axios.get(url, {
```
rande, line 7: Can you check if the sitemap is a local file or a remote URL? So we can pass a local file to limit the number of urls to parse.

percy/sitemap.js
```diff
@@ -0,0 +1,37 @@
...
+
+
+(async() => {
+
+  console.log('[INFO] Fetching sitemap');
+  const pageReferences = await loadSitemap(process.env.SITEMAP_URL)
+
+  console.log('[INFO] Found %d page references', pageReferences.length);
+  for (const i in pageReferences) {
+
```
rande, line 24: Can you use a variable to exit the loop after X iteration (like for now to 100 ?) we can use an env variable for that.

percy/sitemap.js
ekinoben: The original example in Percy's docs included it - wait for a known element to be present before taking screenshot. But it seems - in my testing at least - that the screenshot is taken successfully anyway, so I've commented it out. It might be that we need to put it back in, or even add a variable where you can control which page element we should wait for before taking the screenshot.

percy/sitemap.js
ekinoben: Yup, now loads from local file or URL.

percy/sitemap.js
ekinoben: Yup, pass MAX_PAGES env var to limit number of pages snapshotted.

rande: @ekinoben can you default to 100? Thanks

PR #415, opened 2021-05-31
Switching ansible container to a non-templatized dockerfile, to autom…

axelpavageau:
…ate updates through dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,18 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  # Ansible
+  - package-ecosystem: "docker"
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -5,6 +5,18 @@ updates:
...
       interval: "weekly"
+  # Ansible
+  - package-ecosystem: "docker"
+    directory: "/ansible"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+      - "docker"
+      - "bitcoind"
```
Crow-EH, line 16: mauvais label

axelpavageau: fixed, thanks

PR #422, opened 2021-05-31
restricting dependabot update scope

axelpavageau:


.github/dependabot.yml
```diff
@@ -97,10 +124,13 @@ updates:
       - "ruby"
```
Crow-EH, line 124: You missed one here. Is it on purpose ?

axelpavageau: not at all :)

PR #476, opened 2021-07-26
chore : moving python major tools versions declaration to requirements, to enable dependabots up…

axelpavageau:
…dates


Crow-EH:
It means we'll only be able to maintain one version of azure and ansible at a time.

@axelpavageau are you OK with that ?

If yes, LGTM. :+1:

axelpavageau:
> It means we'll only be able to maintain one version of azure and ansible at a time.
> 
> @axelpavageau are you OK with that ?
> 
> If yes, LGTM. 👍

Azure is not an issue, basically we always use the latest (which is also why I haven't defined any semver exlusion for this pip dependabot action.
As for Ansible, this could be an issue in the future (in which case we'll probably define requirements in per-version subfolders), but we'll handle that in due time if the case arises (we don't use ansible much these days)

Crow-EH (approved):

PR #484, opened 2021-08-06
Adding a new container flavour for Aquasecurity's Cloudsploit

axelpavageau:


cloudsploit/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+LABEL maintainer="Axel Pavageau <axel.pavageau@ekino.com>"
+
+ADD package.json .
+
+RUN apt-get update -qq \
+    && pip install -U pip \
+    && pip install pipenv awscli boto3 \
+    && apt install -y npm git wget zip
+
+RUN npm install \
```
mremi, line 12:
```
npm install -g cloudsploit@^2.0
```
does not work?

Crow-EH: I think @axelpavageau did that to eventually let dependabot scan package.json and upgrade the dependency automatically.

If so, how about setting it up now @axelpavageau ?

axelpavageau: @Crow-EH is right, this is setup for dependabot to handle (although I haven't tested whether it's able to handle this specific dependency git/tag format).

I've added the missing dependabot configurations

.github/dependabot.yml
```diff
@@ -98,6 +98,29 @@ updates:
...
+    directory: "/cloudsploit"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+      - "npm"
+      - "cloudsploit"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
```
Crow-EH, line 123: shouldn't it be `semver-minor` since you put version to `2` in `cloudsploit/config.yml` ?

axelpavageau: it's an ignore statement, so we'll only get minor & patches

Crow-EH: Oooh my bad. LGTM then

Crow-EH (approved):

Crow-EH:
> although I haven't tested whether it's able to handle this specific dependency git/tag format
It should work.
We'll be able to check if it does after merging  & dependabot's next checks : https://github.com/ekino/docker-buildbox/network/updates

PR #514, opened 2021-08-30
build(docker): remove templating and jinja

Crow-EH:
This removes the jinja and templating altogether to enforce having the versions directly in the Dockerfiles, allowing dependabot to update them.

Images that used a `Dockerfile.j2` template now have 1 Dockerfile / version, in a version specific subdirectory (`<image>/<version>/Dockerfile`) + a common `run.sh`.

The subdirectory name is deduced automatically from the version. If no subdirectory exists for an image version, we use `<image>/Dockerfile`.

In any case, the build context will be the main image directory, so we can copy `<image>/run.sh` while building `<image>/<version>/Dockerfile`.

Since we use one folder per version, we can workaround dependabot's limitations by setting one rule per subdirectory.

Also :
* Removes the `dockerfile_dir` parameter: It didn't make any sense anymore now that we are strongly linked to the image folder names (auto matrix, dependabot rules...) while it allowed to put the Dockerfile completely out of the folder.
* Adds missing dependabot docker rules

WDYT @axelpavageau @clemstoquart ?


golang/1.16.Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM golang:1.16.7
+LABEL maintainer="Raphaël Benitte <raphael.benitte@ekino.com>"
```
Crow-EH, line 2: Who's the new golang maintainer ?

axelpavageau: i'd say @rande ?

Crow-EH:
As discussed with @axelpavageau after his tests, I'll need to :
* Use 1 directory / Dockerfile : dependabot doesn't work as intended with multiple dockerfiles
* Use 1 dependabot rule / Dockerfile : //
* Keep modifications of .github/dependabot.yml from procing the whole matrix

clemstoquart (approved):
That's a good idea to remove these templates.

golang/1.16.Dockerfile
Crow-EH: done

rande:
It is very hard to see the gain here. I like to be able to do this can of PR: https://github.com/ekino/docker-buildbox/pull/527/files

But not my call ;)

axelpavageau:
> It is very hard to see the gain here. I like to be able to do this can of PR: https://github.com/ekino/docker-buildbox/pull/527/files
> 
> But not my call ;)

This PR adds dependabot automated PRs for all images flavors in the repo. Currently golang, java, node, php, pynode, python, tezosqa base images must be updated manually.

As for the ability to define / update the images dependencies in a single PR, I don't think this changes anything : the yaml structure is still the same https://github.com/ekino/docker-buildbox/blob/no-template/node/config.yml

rande:
ok, so `dependabot` does not run the image code, just read the `FROM` directive and create a PR based on the value. Make sense.

axelpavageau:
> ok, so `dependabot` does not run the image code, just read the `FROM` directive and create a PR based on the value. Make sense.

Exactly (see https://github.com/ekino/docker-buildbox/pull/528/files for instance). And since it can only read specific definitions (like a `Dockerfile`, a `package.json`, ...) we're trying to move as much dependency management to these standard files.

axelpavageau (approved):
LGTM

PR #648, opened 2022-01-10
WIP: multi-arch images

ekinoben:
This PR illustrates where I've got to with the buildx approach. It builds multi-arch images locally (although several images have unresolved issues on ARM arch, noted below)

The main issue is testing those images within github actions. As we don't seem to be able to build an image that we can then run locally without pushing it to a repo first, the most viable approach now seems to be to drop buildx, create images for each arch individually, and combine them via a manifest. One challenge with this approach is that the individual dockerfiles now make heavy use of the TARGETARCH env var that buildx exposes, that wouldn't be available if we dropped it.

Some known issues / caveats with individual dockerfiles (there will likely be more):

### Bitcoin

- No longer sets `BITCOIN_URL` env var, which was the binary download URL.

### Golang

- Updated Mock version v1.5.0 -> v1.6.0 (the first version that offers an ARM-compatible build).

- There's no ARM-compatible build for [go-testfixtures](https://github.com/go-testfixtures/testfixtures/releases).

### Kubectl

- There's no `linux/arm64` build for `kubectl`. Using `darwin/arm64` instead.
- There's no ARM-compatible build for [kubeval](https://github.com/instrumenta/kubeval/releases).

### Sonar

- Is pulling zlib from Arch Linux package repo. Have added architecture-based switch to pull it from the ARM fork, but since the image is Alpine-based, can we not simply use the same version from APK?

### Terraform

- There's no `linux/arm64` build for `infracost`. Using `darwin/arm64` instead.


Crow-EH:
@ekinoben Do you think you'll have time to finish this soon, or would you be OK with someone else finishing your work ?

ekinoben:
> @ekinoben Do you think you'll have time to finish this soon, or would you be OK with someone else finishing your work ?

Hi @Crow-EH 

Apologies, I've not been able to spend as much time on this recently as I'd hoped.

I've got it - I think - most of the way there, but would appreciate some help getting it over the line, especially in regards to the github build errors.

Right now, it builds arm64 and amd64 images for all buildbox flavours, spins up a temporary image registry, pushes them to that and runs tests from there.

Main issues at the moment are:

1. Github build error - Pip itself seems to be falling over when installing the build script reqs:
`subprocess.CalledProcessError: Command '('lsb_release', '-a')' returned non-zero exit status 1.`
I'm not sure what the cause is? I've not really touched anything here.

2. All the images seem to build and test successfully on my machine locally, except golang, which can't find `aws` cli when testing:

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "aws": executable file not found in $PATH: unknown.
ERRO[0024] error waiting for container: context canceled
> [Error] Command test failed - The docker command executed was `/usr/local/bin/docker container run --platform linux/amd64 localhost:5000/ci-golang:1.16-latest aws --version`.
```

I'm not seeing any errors in the build stage, and when I build the image manually and run it, aws cli is installed and runs as expected. Perhaps I've missed something ovbious in the golang `run.sh` script but I can't see what?


Other than that, all images seem to build and pass tests locally in both arch flavours, but I wasn't able to find ARM-compatible binaries for key dependecies on several images. I guess we're just going to have to start using these images on ARM machines to see how well they fare.

PR #680, opened 2022-02-22
feat(tezosqa and general workflow)

fleandrei:
A new version 1.0 has been added to tezosqa. This version use the version 0.9.0 of smartpy. We can now set the path of smartpy-cli folder via the SMARTPY_PATH build_args argument. 
tezosqa is now added to .github/dependabot.yml. 
Workflow has slightly changed : Some Docker related parameters (the path of docker .sock socket and the connexion timeout) are now modifiable in the base_config.yml file.


.github/dependabot.yml
```diff
@@ -401,3 +401,28 @@ updates:
...
       - "pip"
       - "terraform"
+  # Tezosqa
+  - package-ecosystem: "docker"
+    directory: "/tezosqa/0.1"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+      - "pip"
```
Crow-EH, line 411: `package-ecosystem: "docker"` so:
```suggestion
      - "docker"
```

tezosqa/1.0/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM python:3.8-slim-buster
+
+ARG SMARTPY_VERSION
+ARG SMARTPY_PATH
+
+ENV SMARTPY_INSTALL_URL https://smartpy.io/releases/${SMARTPY_VERSION}/cli
+
+RUN echo "Install system dependencies for python and pip" && \
+    #echo "path : " && smpat = $(${SMARTPY_PATH}) && mkdir ${smpat} && ls -la ~ && exit 1 &&\
```
Crow-EH, line 9: dead code

tezosqa/1.0/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    echo "Done!" && \
+    echo "Install crypto libs" && \
+    apt-get update -q -y && apt-get install -q -y libsodium-dev libsecp256k1-dev libgmp-dev gcc git g++ make && \
+    echo "Done!" && \
+    echo "Install SmartPyDev" && \
+    apt-get install -q -y curl && \
+    curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
+    apt-get install -y nodejs && \
+    curl -s "${SMARTPY_INSTALL_URL}/install.sh" > install.sh && bash install.sh --prefix ${SMARTPY_PATH} --from ${SMARTPY_INSTALL_URL} --yes && rm install.sh && \
+    #ls -la / && ls -la ${SMARTPY_PATH} && exit 1 && \
```
Crow-EH, line 23: dead code

tezosqa/1.0/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    pip install pipenv && \
+    echo "Done!" && \
+    echo "Install crypto libs" && \
+    apt-get update -q -y && apt-get install -q -y libsodium-dev libsecp256k1-dev libgmp-dev gcc git g++ make && \
+    echo "Done!" && \
+    echo "Install SmartPyDev" && \
+    apt-get install -q -y curl && \
+    curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
+    apt-get install -y nodejs && \
+    curl -s "${SMARTPY_INSTALL_URL}/install.sh" > install.sh && bash install.sh --prefix ${SMARTPY_PATH} --from ${SMARTPY_INSTALL_URL} --yes && rm install.sh && \
```
Crow-EH, line 22: You don't need to rm anything if you don't create a file in the first place, check if this works:
```suggestion
     bash <(curl -s "${SMARTPY_INSTALL_URL}/install.sh") --prefix ${SMARTPY_PATH} --from ${SMARTPY_INSTALL_URL} --yes && \
```

.github/dependabot.yml
```diff
@@ -401,3 +401,28 @@ updates:
...
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major", "version-update:semver-minor"]
+  - package-ecosystem: "docker"
+    directory: "/tezosqa/1.0"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+      - "pip"
```
Crow-EH, line 422: docker

.github/dependabot.yml
```diff
@@ -401,3 +401,28 @@ updates:
...
+      - "tezosqa"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major", "version-update:semver-minor"]
+
+
```
Crow-EH, line 428: There should be only one (empty) newline at the end.

base_config.yml
```diff
@@ -1,8 +1,11 @@
 ---
 docker_hub_namespace: &docker_hub_namespace ekino
+DOCKER_SOCK_PATH: &DOCKER_SOCK_PATH unix://var/run/docker.sock
```
Crow-EH, line 3: you don't need to declare an anchor (`&DOCKER_SOCK_PATH`) if you don't use the reference anywhere else in a yaml file (`*DOCKER_SOCK_PATH`).

base_config.yml
```diff
@@ -1,8 +1,11 @@
...
+DOCKER_TIMEOUT: &DOCKER_TIMEOUT 600
```
Crow-EH, line 4: Same thing, the anchor isn't used anywhere

Crow-EH (changes requested):
Thank you for you contribution ! Here are some aditionnal comments :

# The root Dockerfile isn't used anymore
You shouldn't need the root Dockerfile anymore since you made a directory for each version, so please delete `tezosqa/Dockerfile`.

# DRY
It seems like, excluding dead code, only smartpy install is different: That's an oportunity to not repeat yourself. 😄 

You should consider creating a common script in the parent directory with a simple `if`. Look at PHP images  and their `run.sh` for reference.

You just need to `COPY run.sh` as if it was in the same directory as your Dockerfile, we set up the the parent directory as build context for this use case. 🙂

# Wording and changelog
* Please respect this pattern for your commit `<type>(<optional scope>): <description>`. Suggestion: `feat(tezosqua): add version 1.0 and expose docker variable`.
* Please update `CHANGELOG.md`. Just add a new section at the start with a partial date `2022-03`

base_config.yml
```diff
@@ -1,8 +1,11 @@
 ---
 docker_hub_namespace: &docker_hub_namespace ekino
```
Crow-EH, line 2: Remove this anchor too while you're at it :p

Bladrak (changes requested):

base_config.yml
```diff
@@ -1,8 +1,12 @@
 ---
-docker_hub_namespace: &docker_hub_namespace ekino
+docker_hub_namespace: ekino
+#DOCKER_SOCK_PATH: unix://var/run/docker.sock
```
Bladrak, line 3: Remove the commented line if it's not used.

.github/dependabot.yml
```diff
@@ -401,3 +401,26 @@ updates:
       - "dependencies"
       - "pip"
       - "terraform"
+  # Tezosqa
+  - package-ecosystem: "docker"
```
Bladrak, line 405: We don't need to keep 2 versions, remove the 0.1 one and keep the 1.0 one.

Bladrak (changes requested):

.github/dependabot.yml
Crow-EH: ah then the script and the subdirectories are useless and you can go back to a single root Dockerfile if you want @fleandrei , sorry. Let's say you now know how to it. 😄

base_config.yml
```diff
@@ -1,8 +1,12 @@
...
+DOCKER_SOCK_PATH: unix://Users/andrei.fleiser/docker.sock
```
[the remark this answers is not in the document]
Crow-EH: I'm not entirely sure that would be `unix://var/run/docker.sock` on the CI (and it needs to), but I guess you can test.

Othewise just stick to `unix://var/run/docker.sock` or, if you want to and have some time, modify the python script to use the `DOCKER_HOST` env variable (should be `unix://Users/andrei.fleiser/docker.sock` on your device) if it exists, or `DOCKER_SOCK_PATH` if it doesn't.

tezosqa/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    && echo "Install Pytezos" \
+    && apt-get install -q -y pkg-config \
+    && pip install pytezos pytest ipython fire pysodium secp256k1 'fastecdsa<2.0.0' \
+    && echo "Done!" \
+    && echo "Cleanup" \
+    && apt-get -q -y autoremove \
+    && apt-get -q -y clean && apt-get -q -y purge \
+    && rm -rf /var/lib/apt/lists/* /var/lib/dpkg/*-old \
+    && echo "Done!" 
+
```
Bladrak, line 38: Extra blank line to remove

Bladrak (changes requested):

Bladrak (approved):

Bladrak:
Some conflicts to fix, you should rebase it. And maybe squash it so there's only one commit :)

Bladrak (approved):

PR #686, opened 2022-03-04
add terraspace to teraform dockerfile

FridMehdi:
the aim of this pr is to add terraspace project to terrafom image.


terraform/Dockerfile
```diff
@@ -29,3 +29,8 @@ RUN wget https://github.com/infracost/infracost/releases/download/v${INFRACOST_V
 
 RUN wget https://raw.githubusercontent.com/infracost/infracost/v${INFRACOST_VERSION}/scripts/ci/diff.sh -O /opt/diff.sh \
     && chmod +x /opt/diff.sh
+
+RUN echo "deb https://apt.boltops.com stable main" > /etc/apt/sources.list.d/boltops.list \
+    && curl -s https://apt.boltops.com/boltops-key.public | apt-key add - \
```
axelpavageau, line 34: could you run this before the first apt-get update, to avoid doing 2 apt-get updates

terraform/Dockerfile
```diff
@@ -29,3 +29,8 @@ RUN wget https://github.com/infracost/infracost/releases/download/v${INFRACOST_V
...
+    && apt-get update \
+    && apt-get install -y terraspace
```
axelpavageau, line 36: shouldn't we specify a fixed version ?

axelpavageau:
You'll need to update the changelog file as well

terraform/Dockerfile
FridMehdi: @axelpavageau "could you run this before the first apt-get update, to avoid doing 2 apt-get updates" => I s not possible with the current structure of the dockerFile I need` gnupg` to be installed before  adding the mirrors of terraspace to apt.

terraform/Dockerfile
FridMehdi: done version used 1.1.5-1

terraform/Dockerfile
FridMehdi: done in the last commit we used wget instead of curl

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Versions
 ========
 
+2022-03-10
```
axelpavageau, line 4: this should be `2022-03-31`. We release monthly

axelpavageau:
I'm closing this PR as we won't be using terraspace on our projects in the foreseeable future

PR #736, opened 2022-03-24
feat: update awscli for php image

swve:


swve:
status : awscli will be removed from this image in a future update, probably switching to a debian as a base image

PR #849, opened 2022-07-28
[Dind] Feat : build arm64 docker image

Bonnai2112:
**Improvements**
* Dind : build arm64 docker image

**Issue**
* aws: not found on arm64 
* The weird behavior that the install seems to fail at line 78 of the install script, but the install script seems to complete successfully, only to then fail when we try to actually run the installed bin. The bug discussion forum https://github.com/aws/aws-cli/issues/4971

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Versions
 ========
 
+2023-07-31
+----------
+* Dind : build arm64 docker image
+
 2023-06-30
 ----------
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 Versions
 ========
 
+2023-07-30
```
axelpavageau, line 4: 2023-07-31

PR #881, opened 2022-09-02
[WIP][Kubectl] feat : build arm64 docker images

Bonnai2112:
**Improvements**
Kubectl : build arm64 docker image

**Issue**
aws: not found on arm64 in alpine
The weird behavior that the install seems to fail at line 78 of the install script, but the install script seems to complete successfully, only to then fail when we try to actually run the installed bin. The bug discussion forum https://github.com/aws/aws-cli/issues/4971

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Versions
 ========
 
+2022-09-31
+----------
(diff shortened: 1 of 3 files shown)


axelpavageau:
I'm closing this PR as we no longer build this image.

PR #906, opened 2022-09-26
Java : adding java 18 flavor

axelpavageau:


clemstoquart (changes requested):
I'm not sure that's relevant to add a JDK 18 image now that the JDK 19 has been released.

We should probably move to the JDK 19 as soon as the base image is available. What do you think ?

PR #918, opened 2022-10-10
chore: removing kubectl image, moving k8s tooling to AWS & Azure images

axelpavageau:


axelpavageau:
Flagging as draft, because the resulting images are far too large. I need to investigate this before we can merge.

clemstoquart (approved):

PR #978, opened 2022-12-05
feat(node): add jq command in node dockerfile

matthieugremaud2:
This adds "jq" command in node image.


mremi:
Please add a new section in CHANGELOG.md `2022-12-31` with your changes

PR #1005, opened 2023-01-09
 add go1.20 image to build

ArnaudBes:
Create Golang 1.20 image 
Update Changelog 
Update README with tips for local testing


ArnaudBes:
For reviewer @Bladrak

golang/1.20/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM golang:1.20-rc as base
```
Bladrak, line 1: Can't we template that? To avoid having one Dockerfile for each version

Bladrak (changes requested):

golang/Dockerfile.template
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.18.9 as base
+FROM golang:${GOLANG_VERSION} as base
```
Bladrak, line 1: You could use an ARG, avoids using a script :)

axelpavageau: the problem with these methods is that dependabot won't be able to create the pull requests to update the dockerfiles.

Bladrak (changes requested):

golang/config.yml
```diff
@@ -9,23 +9,11 @@ test_config: &test_config
     - modd --version
     - rsync --version
     - swagger version
-
+before_script:
```
Bladrak, line 12: Still needed?

Bladrak (changes requested):

Bladrak (approved):

Bladrak (approved):

ArnaudBes:
replace with new pull request https://github.com/ekino/docker-buildbox/pull/1059

PR #1015, opened 2023-01-17
chore(php): add php 8.2 image & remove 7.4

vladsolntsev:
- Add `linux-headers` because of :
[configure: error: rtnetlink.h is required, install the linux-headers package: apk add --update linux-headers](https://github.com/vladsolntsev/docker-buildbox/actions/runs/3940339455/jobs/6741325494#step:7:5753)
that caused xdebug and blackfire to not install properly

- PHP 8.0 doesn't have any official image on Alpine 3.17, and different musl versions are needed on Alpine 3.16 and 3.17. 

- Everything builds for 8.2 but php-cs-fixer that isn't fully ready yet for 8.2 (https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/6749). Set PHP_CS_FIXER_IGNORE_ENV to bypass the warnings.

### DIFF
--- a/php/8.2/Dockerfile
+++ b/php/8.2/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.4.33-cli-alpine3.15 AS base
+FROM php:8.2.1-cli-alpine3.17 AS base
 LABEL maintainer="Rémi Marseille <marseille@ekino.com>"
 
 ARG APCU_VERSION
@@ -28,7 +28,9 @@ ENV COMPOSER_NO_INTERACTION=1 \
     COMPOSER_MEMORY_LIMIT=-1 \
     TERM=xterm \
     LD_PRELOAD="/usr/lib/preloadable_libiconv.so php" \
-    PHP_CPPFLAGS="$PHP_CPPFLAGS -std=c++11"
+    PHP_CPPFLAGS="$PHP_CPPFLAGS -std=c++11" \
+    # TODO: remove this once php-cs-fixer is officially supporting PHP 8.2
+    PHP_CS_FIXER_IGNORE_ENV=1
 
 COPY run.sh .
 RUN sh run.sh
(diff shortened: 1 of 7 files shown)


vladsolntsev:
@brambaud wip

vladsolntsev:
@mremi @brambaud as php-cs-fixer doesn't support php 8.2 yet, would it be better to wait for its support to release the new image, or release it this month without the cs fixer?

php/8.2/Dockerfile
```diff
@@ -28,7 +28,8 @@ ENV COMPOSER_NO_INTERACTION=1 \
     COMPOSER_MEMORY_LIMIT=-1 \
     TERM=xterm \
     LD_PRELOAD="/usr/lib/preloadable_libiconv.so php" \
-    PHP_CPPFLAGS="$PHP_CPPFLAGS -std=c++11"
+    PHP_CPPFLAGS="$PHP_CPPFLAGS -std=c++11" \
+    PHP_CS_FIXER_IGNORE_ENV=1
```
brambaud, line 32: question: should we add a todo to remove this when php-cs-fixer will officially support php 8.2?

brambaud:
> @mremi @brambaud as php-cs-fixer doesn't support php 8.2 yet, would it be better to wait for its support to release the new image, or release it this month without the cs fixer?

As seen together i think it safe to use `PHP_CS_FIXER_IGNORE_ENV=1` for the moment.
It will allow us to help php-cs-fixer to support PHP 8.2 if we encounter an issue.

Maybe we could add a warning in the changelog explaining that people have to be cautious when using php-cs-fixer with PHP 8.2.

vladsolntsev:
Thanks for the tips on the PHP_CS_FIXER_IGNORE_ENV @brambaud ! To be sure not to forget it in the future, I'll put a todo in the Dockerfile & mention it in the changelog.

php/8.2/Dockerfile
vladsolntsev: I think it's always better to be fully safe so we don't leave it for too long.

brambaud:
LGTM!

PR #1085, opened 2023-05-12
Fix tezosqa

fleandrei:
- Update smartpy and ligo installation process.
- Change the base image from python:3.8-slim-bullseye to docker:20.10.24-dind in order to be able to run docker inside the container. This is due to the fact that the new smartpy executable now needs docker to work. 
- Change apt by apk in order to be compatible with the new base image
- In the config.yml: change test configs, remove all env variables and configure the tezosqa image as being compatible with both amd64 and arm64 architectures

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Versions
 2023-05-31
 ----------
 * PHP: bump musl-dev version
+* Tezosqa : version 2.0 : Update smartpy and ligo installation process, switch base image from python:3.8-slim-bullseye to docker:20.10.24-dind, add compatibility with arm64
 
 2023-04-30
 ----------
--- a/tezosqa/Dockerfile
+++ b/tezosqa/Dockerfile
@@ -1,37 +1,27 @@
-FROM python:3.8-slim-bullseye
+FROM docker:20.10.24-dind 
 
-ARG SMARTPY_VERSION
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 Versions
 ========
+2023-05-12
```
axelpavageau, line 3: could you remove this line ?

tezosqa/Dockerfile
```diff
@@ -1,37 +1,28 @@
...
     && echo "Done!" \
     && echo "Install SmartPyDev" \
-    && apt-get install -q -y curl \
-    && curl -sL https://deb.nodesource.com/setup_12.x | bash - \
-    && apt-get install -y nodejs \
-    && curl -s "${SMARTPY_INSTALL_URL}/install.sh" | bash /dev/stdin --prefix ${SMARTPY_PATH} --from ${SMARTPY_INSTALL_URL} --yes \
-    && echo "export PATH=${SMARTPY_PATH}:\$PATH" >> /etc/profile \
+    && wget https://smartpy.io/smartpy \
+    && chmod +x smartpy && mv smartpy /usr/local/bin \
+    && wget smartpy.io/templates/welcome.py\
```
Bonnai2112, line 17: Not necessary to download this file or execute the test with it into this Dockerfile.

PR #1253, opened 2023-11-30
PHP : adding 8.3 image

axelpavageau:


mremi:
👍 thx

php/config.yml
```diff
@@ -23,24 +23,26 @@ build_args: &build_args
...
     test_config: *test_config
-  "8.2":
+  "8.3":
     platforms: *platforms
     build_args:
       <<: *build_args
       ICONV_VERSION: 1.17-r1
-      MUSL_VERSION: 1.2.4-r2
+      MUSL_VERSION: 1.2.4-r4
+      XDEBUG_VERSION: 3.3.0
```
mremi, line 46: 3.2.0 is ok

mremi: my bad, 3.2.2 is for PHP 8.2 only, have to use 3.3.1 for PHP 8.3

PR #1262, opened 2023-12-06
java: moving installation logic from .sh to native docker instructio…

axelpavageau:
…ns to allow faster failures on build


java/11/Dockerfile
```diff
@@ -10,5 +10,31 @@ FROM base as base-arm64
...
 FROM base-${TARGETARCH}
-COPY run.sh .
-RUN sh run.sh
+
+# Updating packages using sources
+RUN cat /etc/apt/sources.list && \
+    apt-get -qq clean -qq && apt-get -qq update && \
+
+# Install base
+    apt-get -qq -y install build-essential curl git make openssh-client jq unzip groff less && \
```
fpavageau, line 19: It also comes from `run.sh`, but it's unclear why we need to install all those packages, to then remove a bunch of other packages (some of which probably come from installing `build-essential`); I wonder if it wasn't necessary to install awscli v1 via pip.

I guess I'll try to open another PR later on to clean that, as on a `ubuntu:22.04` image it ends up installing 157 more packages, including fonts, ghostscript, imagemagick and lots of libraries.

java/11/Dockerfile
```diff
@@ -10,5 +10,31 @@ FROM base as base-arm64
...
+    unzip -q awscliv2.zip && \
+    ./aws/install && \
+    rm -f awscliv2.zip && \
+    rm -rf aws && \
+
+# Install graphviz
+    apt-get -qq -y install graphviz && \
+
+# Install postgresql-client
+    apt-get -qq -y install postgresql-client && \
```
fpavageau, line 32: Maybe installing `graphviz` and `postgresql-client` with the same command would be slightly faster, since it won't read the dpkg database twice. It could even be grouped with the first install command.

That can be done in another PR, though.

PR #1673, opened 2025-07-13
Add postgresql-client into python 3.13 image

qsomazzi:


axelpavageau:
could you add a test here https://github.com/ekino/docker-buildbox/blob/master/python/config.yml#L10 ?

mremi:
> Hello @axelpavageau i can't really test it, i just make it available in python.

Just add `psql --version`

qsomazzi:
> @qsomazzi could you please rebase?

Done

PR #1726, opened 2025-10-07
Java: added support for Java 25 LTS

axelpavageau:


java/25/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+FROM base AS base-arm64
+ARG AWSCLI_ARCH="linux-aarch64"
+
+FROM base-${TARGETARCH}
+
+# Updating packages using sources
+RUN cat /etc/apt/sources.list && \
+    apt-get -qq clean -qq && apt-get -qq update && \
+    # Install base
+    apt-get -qq -y install build-essential curl git make openssh-client jq unzip groff less && \
```
fpavageau, line 21: Since it's a new image, we could try to trim the packages a bit: I think we can get rid of `build-essential`, `make`, even `groff` and `less` if it's not used interactively.

We could also group the `graphviz` and `postgresql-client` installs with this one.

fpavageau:
LGTM 👍

PR #1892, opened 2026-04-13
feat: add Dockerfile and config for ci-gcp image with gcloud CLI

dieutoutpuissant:


axelpavageau:
Could you please 

- add an entry in https://github.com/ekino/docker-buildbox/blob/master/README.md#available-images to describe the image
- add an entry in https://github.com/ekino/docker-buildbox/blob/master/CHANGELOG.md to explain the changes (add a 2026-04-30 section if it's not there yet)

